PR #32, opened 2018-10-29
update useAsync to use async/await

CarsonMcKinstry:
Since React is starting to move more and more toward cutting-edge JavaScript, I figured this might be one of those cases where async/await would be useful.

### DIFF
--- a/src/useAsync.ts
+++ b/src/useAsync.ts
@@ -12,26 +12,25 @@ const useAsync = <T>(fn: () => Promise<T>, args?) => {
   });
   const memoized = useCallback(fn, args);
 
-  useEffect(() => {
+  useEffect(async () => {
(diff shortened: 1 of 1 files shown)


src/useAsync.ts
```diff
@@ -12,26 +12,25 @@ const useAsync = <T>(fn: () => Promise<T>, args?) => {
   });
   const memoized = useCallback(fn, args);
 
-  useEffect(() => {
+  useEffect(async () => {
```
streamich: I don't think `useEffect` can be an `async` function, because it has to return the unsubscription function below, or can it be async?

cloudever:
Bad idea because @streamich says true and we should pull async await state machine from ts to compile to es5

PR #82, opened 2018-12-18
feat: added useDebounce hook

Jivings:
React hook that delays invoking a function until after wait milliseconds have elapsed since the last time the debounced function was invoked.


Jivings:
Oops, yes true! I added it to *Side-effects*, is that ok?

PR #84, opened 2018-12-26
feat: 🎸add useKeyPress hook

ythecombinator:
First of all, thanks for the amazing project 🚀

This PR basically adds a hook for detecting when the user is pressing a specific key on their keyboard – pretty much like this [sample](https://usehooks.com/#useKeyPress) here does. The key difference it also allows the so called key combos to be set – e.g. firing something specifically when a very weird combination of keys is pressed.

Turns out that handling multiple key presses at once [can grow a lot of code](https://stackoverflow.com/a/12444641) – so I used a [very popular alternative for that](https://github.com/RobertWHurst/KeyboardJS).

Then comes the tradeoff of adding an extra dependency and growing your bundle. I handle this by dynamically importing *KeyboardJS* only if a `useKeyboardJS` option is set true when the hook is called – so, no extra KBs on the bundle for simple use cases.

Screen capture on docs + working sample available [here](https://cl.ly/a5571d1445ad).


src/useKeyPress.ts
```diff
@@ -0,0 +1,63 @@
...
+    }
+    return () => {
+      if (useKeyboardJS) {
+        keyboardjs.unbind(targetKey, customDownHandler, customUpHandler);
+      } else {
+        window.removeEventListener("keydown", regularDownHandler);
+        window.removeEventListener("keyup", regularUpHandler);
+      }
+    };
+  }, []);
```
streamich: The `[]` should probably be

```js
[targetKey, useKeyboardJS]
```

ythecombinator: Sure! My bad this perf thing when not skipping these. I better do `[targetKey, config]`, don't you think so?

streamich: If `useEffect` supports an object, then yes.

streamich: `...Object.values(config)` does not guarantee the order of the values, right? But the order is important here.

j-f1: According to the spec, [`Object.values`](https://read262.netlify.com/fundamental-objects/object-objects#sec-object.values) calls [`EnumerableOwnPropertyNames`](https://read262.netlify.com/abstract-operations/operations-on-objects#sec-enumerableownpropertynames), which reads [`[[OwnPropertyKeys]]`](https://read262.netlify.com/ordinary-and-exotic-objects-behaviours/ordinary-object-internal-methods-and-internal-slots#sec-ordinary-object-internal-methods-and-internal-slots-ownpropertykeys), which calls [`OrdinaryOwnPropertyKeys`](https://read262.netlify.com/ordinary-and-exotic-objects-behaviours/ordinary-object-internal-methods-and-internal-slots#sec-ordinaryownpropertykeys), which specifies an order for the returned array.

ythecombinator: I think the shallow comparison thing could be solved using `useRef` as suggested [here](https://github.com/kentcdodds/react-github-profile/blob/egghead-2018/refactor-03/src/screens/user/components/query.js#L19):

```ts
  useEffect(() => {
    if (isEqual(previousInputs.current, [targetKey, config])) {
      return;
    }

    if (useKeyboardJS) {
      keyboardjs.bind(targetKey, customDownHandler, customUpHandler);
    } else {
      window.addEventListener("keydown", regularDownHandler);
      window.addEventListener("keyup", regularUpHandler);
    }
    return () => {
      if (useKeyboardJS) {
        keyboardjs.unbind(targetKey, customDownHandler, customUpHandler);
      } else {
        window.removeEventListener("keydown", regularDownHandler);
        window.removeEventListener("keyup", regularUpHandler);
      }
    };
  });

  const previousInputs = useRef(null as unknown);
  useEffect(() => {
    previousInputs.current = [targetKey, config];
  });
``` 

But I don't really think it results in the expected behaviour - neither does `[targetKey, useKeyboardJS]`. 

I was reconsidering and I think that, in this case, the second argument to `useEffect` should actually be an empty array. Could you elaborate more on your previous suggestion, @streamich ?

streamich: What do you mean `[targetKey, useKeyboardJS]` does not result in expected behaviour?

I think you should either use that or sort the keys manually:

```js
[targetKey, ...config.entries().sort(([k1], [k2]) => k1 < k2 ? 1 : -1).map(([_, value]) => value)]
```

ythecombinator: nm, @streamich; my bad. 

Anyway, added and it seems to be okay. Could you check again and tell me if there's anything else missing in the PR?

streamich: Looks good.

ythecombinator:
Thanks 😄

PR #99, opened 2019-02-07
Add updateAt method for useList

oakfang:


oakfang:
The best I could find is Ramda's `update` method (https://ramdajs.com/docs/#update). I like `updateAt` better because `update` sounds too much like `set`, which performs a different action.
WDYT?

streamich:
> ...WDYT?

I don't know a better name, unfortunately.

PR #109, opened 2019-02-18
add error state to useGeolocation state

mohamedmagdy17593:
just add error state to useGeolocation.
in case if the user blocks geolocation message.

### DIFF
--- a/src/useGeolocation.ts
+++ b/src/useGeolocation.ts
@@ -1,6 +1,7 @@
 import {useState, useEffect} from 'react';
 
 export interface GeoLocationSensorState {
+  loading: boolean,
   accuracy: number,
   altitude: number,
   altitudeAccuracy: number,
@@ -8,11 +9,13 @@ export interface GeoLocationSensorState {
   latitude: number,
   longitude: number,
   speed: number,
-  timestamp: number
+  timestamp: number,
+  error?: Error | PositionError
 }
 
 const useGeolocation = () => {
   const [state, setState] = useState({
+    loading: true,
     accuracy: null,
     altitude: null,
     altitudeAccuracy: null,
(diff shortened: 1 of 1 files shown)


src/useGeolocation.ts
```diff
@@ -8,7 +8,8 @@ export interface GeoLocationSensorState {
   latitude: number,
   longitude: number,
   speed: number,
-  timestamp: number
+  timestamp: number,
+  error: boolean
```
streamich: Do you think we can store the error here, or `boolean` is enough?

```ts
error?: Error | PositionError,
```

mohamedmagdy17593: Yes, you are right.
What about using `success` and `error` object so it can be used like that
```javascript
const {success, error} = useGeolocation()
```

streamich: Maybe call it `loading`, for consistency with other libs; and remove the default `null`s:

image

streamich:
```js
const [state, setState] = useState({
  loading: true,
  timestamp: Date.now(),
});
```

mohamedmagdy17593: So it will be something like ?
```javascript
const [state, setState] = useState({
  loading: true,
  timestamp: Date.now(),
  error?: PositionError,
  data?: GeolocationPosition, // or info 
})
```

mohamedmagdy17593: that will be a breaking change ?

streamich: > that will be a breaking change ?

Actually, maybe better not to add a breaking change and just add `loading` and `error` fields.

mohamedmagdy17593: Ok. that's right.

mohamedmagdy17593: I don't have access to my labtop now, So i will do it tomorrow

mohamedmagdy17593:
Ok, you can check now.

PR #139, opened 2019-03-10
Add useKeyPressEvent custom hook with associated doc

google-mac:
This focuses on utilizing the onKeyup and/or onKeydown events unlike useKeyPress which involves state. useKeyPressEvent uses useKeyPress to achieve this functionality.

My needs were to fire an event on keyup and/or keydown, not just know whether it was pressed or not.

I am partial on the parameter implementation. I like named parameters when there are more than 2, especially when some are optional, but most custom hooks I have seen do not use objects. Let me know your thoughts on this. Also, as you'll see in the doc, this code sets a precedence on the keydown functionality (I assume this is used more), but maybe having named parameters would be better to easily do one or the other just as easily.

Lastly, I am automatically determining whether to use the useKeyboardJS param based on the length of the "key" passed. I'm honestly not too familiar with that package, so if this will cause issues I can change the API.

### DIFF
--- a/docs/useKeyPressEvent.md
+++ b/docs/useKeyPressEvent.md
@@ -1,0 +1,57 @@
+# `useKeyPressEvent`
+
+React UI sensor hook that detects when the user is pressing a specific
+key on their keyboard and fires a specified keyup and/or keydown effect. If
+you only need to retrieve the state, see [useKeyPress](#).
+
+Complex bindings like detecting when multiple keys are held down at the same
+time or requiring them to be held down in a specified order are also available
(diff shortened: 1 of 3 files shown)


google-mac:
This could possibly be named `useKeyPressEffect` to denote that it works like an effect.

google-mac:
@streamich or whomever is maintaining this project, could I get some feedback? Thank you! 😁

streamich:
@google-mac Thanks, that a great hook!

Sorry, I was actually reviewing your PR 10 days ago but got distracted.

I've added few missing things and merged it in #149

PR #145, opened 2019-03-16
useAsyncRetry

lvl99:
I've done a first pass at adding a retry/refresh ability for `useAsync`.

I know I've potentially done some debatable things. I'm just curious to get peoples' feedback and thoughts and any extra guidance on whether this is a viable feature for this library.

I've been using `useAsync` to load in select field options, however when network/server errors get in the way, I wasn't sure how to refresh the async call again, so I thought maybe it'd be a good idea to have this ability on the `useAsync` hook. Other use-cases I can think of are for refreshing a table's real-time data, an activity feed, etc.

Additionally it'd be nice to work in a `cancel` feature too...

(P.S. Ignore any formatting differences for the moment, I'll fix those up and disable prettier later)


streamich:
This is an interesting idea, do you think it is possible make this hook `useAsyncWithRetry`, which internally uses `useAsync`?

lvl99:
I wasn't sure how to call the useAsync hook again, hence why I baked in the retry and some extra state management into a same-but-different separate hook, but I guess I could just curry the async fn with an attempt param which would force the useAsync to fire again (newly generated curried fn wouldn't be memoised). I'll have a play!

lvl99:
@streamich I created a new `useAsyncRetry` hook that leverages `useAsync`. Let me know what you think 👍

src/useAsyncRetry.ts
```diff
@@ -0,0 +1,19 @@
+import { useCallback, useState } from 'react';
+import useAsync from './useAsync';
+
+const useAsyncRetry = <T>(fn: () => Promise<T>, args: any[] = []) => {
+  const [attempt, setAttempt] = useState<number>(0);
+  const memoized = useCallback(async () => await fn(), [...args, attempt]);
```
streamich: Can this be written without `async/await` an extra function wrapping?

```js
const memoized = useCallback(fn, [...args, attempt]);
```

src/useAsyncRetry.ts
```diff
@@ -0,0 +1,19 @@
...
+  const state = useAsync(memoized);
+
+  const retry = useCallback(() => {
+    if (state.loading) {
+      return;
```
streamich: Shall we warn user in development mode that she is trying to "retry" while promise is being resolved?

```js
if (state.loading) {
  if (process.env.NODE_ENV !== 'production') {
    console.warn('You are calling useAsyncRetry hook retry() method while loading in progress, this is a no-op.');
  }
  return;
}
``

File: src/useAsyncRetry.ts (hunk 1 above)
Comment (lvl99) [reply src/useAsyncRetry.ts]: I tried this first (it seemed more intuitive to me) but for some reason it wouldn't work on the storybook story, due to the original async function having already resolved. Wrapping it in another async seemed to make it work as advertised.

File: src/useAsyncRetry.ts (hunk 2 above)
Comment (lvl99) [reply src/useAsyncRetry.ts]: Yep, I had a console message originally for debugging purposes. I think I'm more inclined to do a `process.env.NODE_ENV === "development"` or to use `console.log` to avoid aborting build pipelines (CRA aborts pipeline when `CI=true` and any `console.warn` or `console.error` messages are outputted).

File: src/useAsyncRetry.ts (hunk 1 above)
Comment (lvl99) [reply src/useAsyncRetry.ts]: I actually don't think there's a need to memoise the `fn` with `useCallback` if there's no need to manipulate it (since it's a function it will always be passed by reference anyway). I think by creating an async factory wrapper around it, it's probably more likely to have a new async state since we're generating a new async function (and therefore will need memoisation with `useCallback`).

Comment (lvl99) [reply src/useAsyncRetry.ts]: Now that I think about it even further, it could even just be simplified to `useCallback(() => fn(), [...args, attempt])` since the return value of `fn` will always be a promise anyway.

File: src/useAsyncRetry.ts [hunk 3]
```diff
@@ -8,6 +8,9 @@ const useAsyncRetry = <T>(fn: () => Promise<T>, args: any[] = []) => {
 
   const retry = useCallback(() => {
     if (state.loading) {
+      if (process.env.NODE_ENV !== 'development') {
```
streamich: `===`

lvl99: Haha, whoops!

lvl99: Updated 👍

lvl99:
Boom 💥 🎉

PR #162, opened 2019-03-23
useLockBodyScroll

wardoost:
Based on https://usehooks.com/useLockBodyScroll/ but added a `enabled` parameter so you can disable the lock on scrolling with a state toggle

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,7 @@
   - [`useCss`](./docs/useCss.md) &mdash; dynamically adjusts CSS.
   - [`useFavicon`](./docs/useFavicon.md) &mdash; sets favicon of the page.
   - [`useLocalStorage`](./docs/useLocalStorage.md) &mdash; manages a value in `localStorage`.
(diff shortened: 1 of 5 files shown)


wardoost:
Would it make sense to have some state inside the hook? So instead of doing this
```jsx
import {useLockBodyScroll, useToggle} from 'react-use';

const Demo = () => {
  const [locked, toggleLocked] = useToggle(false)
  useLockBodyScroll(locked);

  return (
    <div>
      <button onClick={toggleLocked}>
        {locked ? 'Unlock' : 'Lock'}
      </button>
    </div>
  );
};
```
you could do this
```jsx
import {useLockBodyScroll} from 'react-use';

const Demo = () => {
  const [locked, setLocked] = useLockBodyScroll(true);

  return (
    <div>
      <button onClick={toggleLocked}>
        {locked ? 'Unlock' : 'Lock'}
      </button>
    </div>
  );
};
```

streamich:
I like the first controlled version better.

PR #166, opened 2019-03-25
useMouse

### LINKED ISSUE #47: useMouse
Port mouse sensor: https://github.com/streamich/libreact/blob/master/docs/en/MouseSensor.md

wardoost:
Hook that tracks mouse position, ported from the libreact's [`MouseSensor`](https://github.com/streamich/libreact/blob/master/docs/en/MouseSensor.md). Still have to add the `whenHovered` option though. I'm also not a 100% sold on the naming of the returned state variables. I think `posX` and `posY` could just be `x` and `y`?

Closes #47

### DIFF
--- a/src/useMouse.ts
+++ b/src/useMouse.ts
@@ -26,6 +26,7 @@ const useMouse = (ref: RefObject<HTMLElement>): State => {
 
   useEffect(() => {
     const handler = (event: MouseEvent) => {
+      cancelAnimationFrame(frame.current)
       frame.current = requestAnimationFrame(() => {
         if (ref && ref.current) {
           const {left, top} = ref.current.getBoundingClientRect()
@@ -49,11 +50,8 @@ const useMouse = (ref: RefObject<HTMLElement>): State => {
     document.addEventListener('mousemove', handler);
 
     return () => {
-      if (frame.current) {
-        cancelAnimationFrame(frame.current);
-      }
-
-      document.addEventListener('mousemove', handler);
+      cancelAnimationFrame(frame.current);
+      document.removeEventListener('mousemove', handler);
     };
   }, []);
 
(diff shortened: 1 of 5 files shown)


wardoost:
What about a `useWindowMouse` hook that only returns `x` and `y` coords relative to the window? Could also be done when not passing in a ref in this hook

src/useMouse.ts
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    document.addEventListener('mousemove', handler);
+
+    return () => {
+      if (frame.current) {
+        cancelAnimationFrame(frame.current);
+      }
+
+      document.addEventListener('mousemove', handler);
```
streamich: `removeEventListener`

src/useMouse.ts
```diff
@@ -0,0 +1,63 @@
...
+    posY: 0,
+    elX: 0,
+    elY: 0,
+    elH: 0,
+    elW: 0,
+  });
+
+  useEffect(() => {
+    const handler = (event: MouseEvent) => {
+      frame.current = requestAnimationFrame(() => {
```
streamich: Should there be 

```js
cancelAnimationFrame(frame.current)
```

before this line?

streamich:
`useWindowMouse` would be useful, too. I don't know if it is better to put it inside this hook, or make a separate one.

src/useMouse.ts
wardoost: Fixed

src/useMouse.ts
wardoost: Yeah, I think so. I noticed I didn't put in above [this line](https://github.com/streamich/react-use/blob/master/src/useScroll.ts#L19) in `useScroll` either 😱

PR #169, opened 2019-03-25
feat: more precise size for useMouse

ShizukuIchi:
`element.offsetHeight` and `element.offsetWidth` will round the value to an integer.
`width` and `height` from `element.getBoundingClientRect()` is more precise.


wardoost:
It's more than just a difference in precision apparently, from [MDN](https://developer.mozilla.org/en-US/docs/Web/API/CSS_Object_Model/Determining_the_dimensions_of_elements):

> In case of transforms, the offsetWidth and offsetHeight returns the element's layout width and height, while getBoundingClientRect() returns the rendering width and height.

So when applying transforms to the element `elX` and `elY` would be of the rendered element and `elW` and `elH` of the element's layout.

PR #170, opened 2019-03-26
useMouse whenHovered option

wardoost:
Added `whenHovered` option as the second parameter similar to [`MouseSensor`](https://github.com/streamich/libreact/blob/master/docs/en/MouseSensor.md) from libreact. I also made the demo cute.


wardoost:
I added an `enabled` argument to `useHoverDirty` to make it controllable. This means it can be reused for this `whenHovered`. It's the same pattern as the `enabled` option in `useLockBodyScroll`, not sure if it's a good pattern 😬

wardoost:
Maybe a `disable` argument is better than `enabled`?

streamich:
+1 for the emoji demo :)

Issue #38, opened 2018-10-29
Configure ESM build


cloudever:
It would be great to support ESModules to provide better tree shaking, rollup might be helpful

P.S. I mean mjs format

streamich:
Done in https://github.com/streamich/react-use/pull/178

PR #178, opened 2019-03-27
feat: emit code in ES Module format

liuyuchenzh:
Export the whole library in ES Module format as well so that bundlers like webpack, rollup and parcel can enable code splitting when `import { usexxx } from "react-use"`.

> currently only export in commonjs format

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -66,6 +66,7 @@ typings/
 
 # build output
 lib/
+esm
 
 .DS_Store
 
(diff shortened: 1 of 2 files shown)


streamich:
https://github.com/streamich/react-use/issues/38

liuyuchenzh:
Personally prefer `esm`. Shall we shake hands on it?

streamich:
@liuyuchenzh I think `esm/` would be good.

PR #196, opened 2019-03-31
useEffectOnce

wardoost:
A modified [`useEffect`](https://reactjs.org/docs/hooks-reference.html#useeffect) hook that only runs once. Similar to [useLifecycles](https://github.com/streamich/react-use/blob/master/docs/useLifecycles.md) but implementation can be used more similar to React's `useEffect`.

This implementation uses a ref to know if the effect already ran. I have seen a lot of people using this kind of implementation but I'm not sure if it's the best way. A more straightforward way would be:

```ts
import {useEffect, EffectCallback} from 'react';

const useEffectOnce = (effect: EffectCallback) => {
  useEffect(effect, []);
}

export default useEffectOnce;
```

I thought I read somewhere that the ref implementation is preferable but I can't find it back.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,6 +83,7 @@
     <br/>
     <br/>
 - [**Lifecycles**](./docs/Lifecycles.md)
+  - [`useEffectOnce`](./docs/useEffectOnce.md) &mdash; a modified [`useEffect`](https://reactjs.org/docs/hooks-reference.html#useeffect) hook that only runs once.
   - [`useEvent`](./docs/useEvent.md) &mdash; subscribe to events.
   - [`useLifecycles`](./docs/useLifecycles.md) &mdash; calls `mount` and `unmount` callbacks.
   - [`useRefMounted`](./docs/useRefMounted.md) &mdash; tracks if component is mounted.
--- a/docs/useEffectOnce.md
+++ b/docs/useEffectOnce.md
@@ -1,0 +1,27 @@
+# `useEffectOnce`
+
+React lifecycle hook that runs an effect only once.
+
+## Usage
+
+```jsx
+import {useEffectOnce} from 'react-use';
+
+const Demo = () => {
+  useEffectOnce(() => {
+    console.log('Running effect once on mount')
+
+    return () => {
+      console.log('Running clean-up of effect on unmount')
+    }
+  });
+
+  return null;
+};
+```
+
+## Reference
+
+```js
+useEffectOnce(effect: EffectCallback);
+```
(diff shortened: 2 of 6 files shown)


streamich:
Maybe it would be interesting to run effect once per `deps` change.

```js
const useEffectOnce = (effect, deps) => {
  const didRun = useRef(false);

   useEffect(() => {
    if (!didRun.current) {
      didRun.current = true;
      const res = effect();
      return () => {
        didRun.current = false;
        return res();
      };
    }
  }, deps);
}
```

wardoost:
@streamich Isn't running the effect once per deps change is the same as `useEffect`?

wardoost:
I think we could also deprecate the `useLifecycles` hook if we accept this hook as they serve similar purposes. I would prefer this implementation as it's closer to React's `useEffect` so it should be easier to understand when you are already using hooks.

streamich:
> Isn't running the effect once per deps change is the same as `useEffect`?

@wardoost yeah, it is, haven't had coffee yet 😆 

I like the name `useEffectOnce`, better than `useLifecycles`, too. And the fact that you return the unmount function, same as in `useEffect` hook, is nice.

Audit of what we have now:

```js
useMount(mount)
useUnmount(unmount)
useLifecycles(mount, unmount)
useEffectOnce(function mount () {return unmount})
useUpdateEffect(update)
```

`useMount` is kinda nice semantic alias, although if you work with hooks a lot, you don't really have  a problem of doing `useEffect(mount, [])`. I'm kinda leaning on keeping `useMount` and `useUnmount`.

`useLifecycles` seems, indeed, inferior to `useEffectOnce` and can be removed.

`useUpdateEffect` is very useful.

streamich:
I like the 

```js
const useEffectOnce = effect => useEffect(effect, []);
```

implementation much more, unless there are good reasons of using the ref implementation.

wardoost:
I would definitely keep `useMount` and `useUnmount` because they are indeed nice semantic aliases. In `useMount` we might not want to return `mount` and change it to something like this:

```js
const useMount = (mount) => {
  useEffect(() => {
    mount();
  }, []);
};
```

This way you only run your effect on mount which is more true to its name and a way to avoid overlapping functionality. `useMount` and `useUnmount` could actually also use `useEffectOnce` internally.

`useUpdateEffect` is very useful indeed 👍

wardoost:
I pushed the ref-less implementation more too. Let's keep it at that until there actually is a reason that the ref implementation is better.

PR #205, opened 2019-04-02
Update readme + fixing docs

### LINKED ISSUE #183: Parcel build breaks, keyboardjs and rebound missing
I see #167 moved `keyboardjs` and `rebound` to `devDependencies`, but this seems to be causing issues, at least with Parcel which is what I use on my project.

```
$ npx parcel build ./src/index.html --public-url '/'
🚨  /Users/wesley/Sites/subway-time/node_modules/react-use/esm/useKeyPress.js:12:15: Cannot resolve dependency 'keyboardjs'
  10 |     var useKeyboardJS = config.useKeyboardJS;
  11 |     if (useKeyboardJS) {
> 12 |         import("keyboardjs").then(setKeyboardJs);
     |               ^
  13 |     }
  14 |     var regularDownHandler = function (_a) {
  15 |         var key = _a.key;
```

wardoost:
I updated the README and fixed typos and inconsistencies in docs. The current layout looks fun on GitHub but breaks on NPM. This more minimal readme looks ok on both websites. I also added some more info on how to import modules from this library separately. Maybe it would be nice to have a similar header image as [`libreact`](https://github.com/streamich/libreact)?

Closes #183


README.md
```diff
@@ -1,134 +1,124 @@
...
-<br/>
-
-<h2 align="center"><sub>License</sub></h2>
-
-<p align="center">
-  <a href="./LICENSE">Unlicense</a> &mdash; public domain.
-</p>
-
-[img-demo]: https://img.shields.io/badge/demo-%20%20%20%F0%9F%9A%80-green.svg
+# 👍 react-use
```
streamich: I'm not sure I would want to do layout changes in README (though, badges can be added and `babel-plugin-import` possibly, too). The README does look different in NPM, because NPM does not center the text, but the spacing and centering was designed specifically for GitHub, I'm not sure I would want to change it now.

wardoost:
@streamich I have updated the readme according to your suggestions. I added the suggestion about `babel-import-plugin` but feel the layout looks weird because it is all centered. For the header it's fine but for body text it's looks like a bad layout to me.

Other questions:
- Aren't hooks supported from 16.8.0? https://reactjs.org/blog/2019/02/06/react-v16.8.0.html
- What about removing the demo badge in the hooks list? It's inconsistent at the moment and is not really necessary if we link to the storybook page on top.

streamich:
> Aren't hooks supported from 16.8.0?

They are, it's just when I last updated that sentence it was `16.8.1`, also I'm not sure when `useState` behaviour changed but if you use `16.7.*-alpha` or `16.8.0` (don't remember which one) this library won't work due to different `useState` behaviours, I'm just mentioning this because maybe we should not add the reference to the first version of React with hooks, maybe remove it altogether, or simply mention that you have to use React with hooks.

> It's inconsistent at the moment and is not really necessary if we link to the storybook page on top.

Some links send user to CodeSandbox page.

Issue #208, opened 2019-04-03
useCopyToClipboard


scottc11:
Here is the version I am using at work.

```
const useCopyToClipboard = (timeout = null) => {

  const [success, setSuccess] = useState(false);

  const copyToClipboard = (text) => {

    if (typeof text == "string" || typeof text == "number" ) {
      const element = document.createElement('textarea'); // create textarea HTML element
      element.value = text;                               // add the text to be copied to the element
      document.body.appendChild(element);                 // add element to DOM
      element.select();                                   // select the text
      document.execCommand('copy');                       // execute copy command
      document.body.removeChild(element);                 // remove element from DOM
      setSuccess(true);
    }
    else {
      setSuccess(false);
      console.error(`Cannot copy typeof ${typeof text} to clipboard, must be a valid string or number.`);
    }
  }

  useEffect( () => {
    // if timeout given, set success to false
    if (timeout) {
      if (success) setTimeout( () => setSuccess(false), timeout);
    }
  }, [success])

  return [success, copyToClipboard];
}
```
use:
```
const myComp = () => {
  const [copiedToClipboard, copyToClipboard] = useCopyToClipboard();
  const myText = "hello!!!";
  return <div onClick={() => copyToClipboard(myText)}>{myText}</div>
}
```

I only suggest adding it to this library because it is kind of a wonky/hacky way of doing something which I expected to be simple. It also seems to be the standard approach for cross-browser support.

NOTE: I took the actual "copy to clipboard" code from this [HackerNoon Arcticle](https://hackernoon.com/copying-text-to-clipboard-with-javascript-df4d4988697f)  as well as this [stackoverflow  question](https://stackoverflow.com/questions/400212/how-do-i-copy-to-the-clipboard-in-javascript).

wardoost:
Really useful hook 👍

I generally use [`clipboard-polyfill`](https://github.com/lgarron/clipboard-polyfill) for this functionality so we could also reuse that library? Do you want to submit a PR for this @scottc11?

scottc11:
thanks @wardoost, I'll submit a PR.

I looked into the clipboard-polyfill and we could for sure reuse that library, but would that mean we would have to add it as a dependancy?

streamich:
Maybe it makes sense to inject `copy` functionality, so 3rd party library can be included, if needed.

```js
const copyDefault = () => {
      const element = document.createElement('textarea'); // create textarea HTML element
      element.value = text;                               // add the text to be copied to the element
[...]

PR #215, opened 2019-04-08
Improved useCopyToClipboard

wardoost:
The current implementation assumes an asynchronous copy to clipboard function but uses [`copy-to-clipboard`](https://github.com/sudodoki/copy-to-clipboard) by default which is synchronous. `copy-to-clipboard` also [returns a boolean](https://github.com/sudodoki/copy-to-clipboard#api) if additional user interaction was needed as a prompt modal is presented in browsers that don't support `execCommand`. This value is not accessible in the current version. 2 examples why the hook itself has to determine how text is copied to the clipboard as I have pointed out in https://github.com/streamich/react-use/issues/208#issuecomment-480472083.

After playing around with it a little bit more I came to the conclusion that it might be a bad pattern to have callbacks in the parameters of a hook. I changed the hook to return a state object with the `value` that was copied, an `error` when copying fails and a `noUserInteraction` boolean forwarded from `copy-to-clipboard`. You can have side effects depending on the state of `copyToClipboard` by using `useEffect`. An example for @scottc11's timeout effect in #208:

```jsx
const [text, setText] = React.useState('');
const [showSuccess, setShowSuccess] = React.useState(false);
const [state, copyToClipboard] = useCopyToClipboard();

React.useEffect(() => {
  if (state.value) {
    setShowSuccess(true);
    setTimeout( () => setShowSuccess(false), 2000);
  }
}, [state])

return (
  <div>
    <input value={text} onChange={e => setText(e.target.value)} />
    <button type="button" onClick={() => copyToClipboard(text)}>copy text</button>
    {showSuccess && `Copied ${text} to clipboard`}
  </div>
)
```

This feels more like a react hooks way of doing things as the `onSuccess` and `onError` callbacks they actually are effects.

### DIFF
--- a/docs/useCopyToClipboard.md
+++ b/docs/useCopyToClipboard.md
@@ -2,13 +2,23 @@
 
 Copy text to a user's clipboard.
 
-
 ## Usage
-
-Basic usage
 
 ```jsx
 const Demo = () => {
+  const [text, setText] = React.useState('');
+  const [state, copyToClipboard] = useCopyToClipboard();
+
+  return (
+    <div>
+      <input value={text} onChange={e => setText(e.target.value)} />
+      <button type="button" onClick={() => copyToClipboard(text)}>copy text</button>
+      {state.error
+        ? <p>Unable to copy value: {state.error.message}</p>
+        : state.value && <p>Copied {state.value}</p>}
+    </div>
+  )
+
   const [text, setText] = React.useState('');
   const [copied, copyToClipboard] = useCopyToClipboard(text);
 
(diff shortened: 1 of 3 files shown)


streamich:
> The current implementation assumes an asynchronous copy to clipboard function but uses `copy-to-clipboard` by default which is synchronous

That should be fine, because `await` still handles it, and, if it didn't, it could have been transformed.

I like that the copied `value` is stored in state, instead of a boolean; it is much better.

One thing I'm not sure about, is throwing the errors, which should not happen in correctly written app. If they were warnings, wrapped in `.NODE_ENV !== 'production'` they would have been stripped away at build time and have smaller bundle size, so apps don't get penalized. For example, if app is written in TypeScript, the build will already fail if the `value` parameter is not a `string`, so adding a type assertion without `.NODE_ENV !== 'production'` adds this code for TypeScript users for no benefit.

wardoost:
Type errors definitely need to be stripped away in production builds 👍

PR #216, opened 2019-04-08
useAsyncCallback

wardoost:
`useAsync` with a callback

Alternative names: useAsyncManual, useAsyncCall, useAsyncFn

### DIFF
--- a/docs/useAsyncCallback.md
+++ b/docs/useAsyncCallback.md
@@ -1,0 +1,36 @@
+# `useAsyncCallback`
+
+React hook that returns state and a callback for an `async` function or a
+function that returns a promise. The state is of the same shape as `useAsync`.
+
+## Usage
+
+```jsx
+import {useAsyncCallback} from 'react-use';
+
+const Demo = (url) => {
+  const [state, fetch] = useAsyncCallback(async () => {
+    const response = await fetch(url);
+    const result = response.text();
+    return result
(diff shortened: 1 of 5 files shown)


docs/useAsyncCallback.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Usage
+
+```jsx
+import {useAsyncCallback} from 'react-use';
+
+const Demo = (url) => {
+  const [state, fetch] = useAsyncCallback(async () => {
+    const response = await fetch(url);
+    const result = response.text();
```
streamich: Should it be

```js
const result = await response.text();
```

wardoost:
@streamich `useAsync` could definitely reuse this hook, will push that soon 👍 

Would it make sense to rename or alias `useAsync` to `useAsyncEffect`?

streamich:
> Would it make sense to rename or alias `useAsync` to `useAsyncEffect`?

Maybe, or call this hook `useAsyncEffect`, up to you. But better *alias* so it is not a breaking change just because of this naming.

wardoost:
I would not call this hooks `useAsyncEffect` as it's not an effect. Adding unnecessary aliases makes it harder for new users to know which hook they need. I think it's maybe best to rename this hook to `useAsyncFn` to align it with for example `useThrottleFn`.

PR #220, opened 2019-04-10
Add `useBeforeUnload` react hook

ilyalesik:


src/useBeforeUnload.ts
```diff
@@ -0,0 +1,17 @@
+import {useEffect} from "react";
+
+const useBeforeUnload = (message?: string) => {
+  useEffect(() => {
+    window.onbeforeunload = e => {
```
wardoost: Needs `event.preventDefault()`? https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event#Examples

Also, are there any particular reasons to use this type of event handler over `window.addEventListener(...)`?

ilyalesik: Done

streamich:
@wardoost this repo is set up to make releases automatically according to Angular semantic commit convention, commits in master that start with `feat: ...` are automatically released as minor versions, commits that start with `fix: ...` are automatically released as fixes. Other commits are mostly ignored. There are also some other handy commit types: https://github.com/angular/angular/blob/master/CONTRIBUTING.md#type

Issue #101, opened 2019-02-07
Tests


streamich:
- [ ] Start adding unit tests for hooks.
- [ ] What are the best practices for testing React Hooks?

lvl99:
I've been using [`react-hooks-testing-library`](https://github.com/mpeyper/react-hooks-testing-library) which seems pretty good so far.

lvl99:
When I was testing some async hooks I wasn't able to adequately get around the `act()` error messages from generating. I know with `CI=true` on `create-react-app` that these messages can prevent a build pipeline from succeeding, even if the tests themselves passed.

ZWkang:
pick the test tools and make a test branch start it.

ankithkonda:
Please see pull request #221 :) let me know what you guys think

Belco90:
Can I help with this? I love this collection of hooks and I like adding tests so I'm more than happy to help with this.

I've written tests for hooks before just with react-testing-library by creating a fake component within the test itself, but react-hooks-testing-library is another good option (which basically do the same: wrap the hook into a fake component and return the results).

I've seen some tests already added by @ankithkonda so the set up and the structure is already defined, I can add more tests following same pattern.

streamich:
@Belco90 that is very very welcome!

Belco90:
Ok so my idea is to add tests for all hooks section by section on storybook (animation, lifecycle, state, etc) and create a PR for each of those sections. Does that make sense for you? I'll start from animation section since it already has some tests and are simple to be tested.

streamich:
> Does that make sense for you?

@Belco90 Yes, it does.

Belco90:
It took me longer as expected as I was on holidays, but PR for first block of tests ready! https://github.com/streamich/react-use/pull/500

Belco90:
Second block of tests ready! https://github.com/streamich/react-use/pull/554

From now on I'll keep adding tests by blocks but I'll create a PR for each hook test, so 1) I avoid re-doing tests done by someone else while I create my PR, 2) I have my branches up to date and I don't have to keep it updated with master, and 3) I can merge bug fixes asap if I found any issue while testing.

Belco90:
@streamich I've edited your description to use it as pending hooks and blocks I need to test.

gerardkabre:
[...]

PR #221, opened 2019-04-10
Add tests for useWindowSize

ankithkonda:
Referencing issue - #101  

My suggestion would be to create a folder for each hook and have any related file in there, including tests.  

In the spirit of using TS, I've used TS for the test as well. 

Not sure if this is the best practice for testing hooks. let me know what you guys think


wardoost:
Looks like a good start 👍 We could we use [react-hooks-testing-library](https://github.com/mpeyper/react-hooks-testing-library) which will probably allow us to only have 1 test file per hook? I would also put all tests in a `__tests__` folder to align it with the `__stories__` folder.

ankithkonda:
good call! I've updated to __tests__ and the test now uses [react-hooks-testing-library](https://github.com/mpeyper/react-hooks-testing-library) :)

jest.config.js
```diff
@@ -0,0 +1,185 @@
+// For a detailed explanation regarding each configuration property, visit:
```
wardoost: Do we need all these comments? It seems more clear to me to only include used config options

package.json
```diff
@@ -57,15 +60,19 @@
...
     "fork-ts-checker-webpack-plugin": "1.0.1",
     "gh-pages": "2.0.1",
+    "jest": "^24.7.1",
     "keyboardjs": "2.5.1",
     "markdown-loader": "5.0.0",
     "react": "16.8.4",
     "react-dom": "16.8.4",
+    "react-hooks-testing-library": "^0.4.0",
     "react-spring": "6.1.10",
+    "react-testing-library": "^6.1.2",
```
wardoost: Not a direct dependency anymore?

src/__tests__/useWindowSize.test.tsx
```diff
@@ -0,0 +1,57 @@
+import * as React from "react";
```
wardoost: Don't have to import React

src/__tests__/useWindowSize.test.tsx
```diff
@@ -0,0 +1,57 @@
...
+import useWindowSize from "../useWindowSize"
+import { renderHook, cleanup, act } from 'react-hooks-testing-library'
```
wardoost: `cleanup` not used

jest.config.js
ankithkonda: Agreed!

package.json
ankithkonda: Ah forgot to remove this from the first commit

src/__tests__/useWindowSize.test.tsx
ankithkonda: woops forgot this too!

src/__tests__/useWindowSize.test.tsx
ankithkonda: Added cleanup according to the example on their readme

ankithkonda:
All done :)

wardoost:
Looks great, would you add a `test:watch` script to `package.json`?

ankithkonda:
👍done! 👁

wardoost:
🙏

PR #223, opened 2019-04-10
fix(useIdle): include 'ms' prop in uesEffect dependencies

### LINKED ISSUE #222: useIdle should reset on delay change
In `useIdle` hook, currently the events prop is used as dependency array of `useEffect` hook.
The `ms` prop is not included in dependencies which doesn't allow changing delay once component is mounted.

So, instead of
```javascript
...
useEffect(() => {
  ...
}, events)
...
```

it should be
```javascript
...
useEffect(() => {
  ...
}, [ms, events])
...
```

ManojBahuguna:
Fixes #222

### DIFF
--- a/src/useIdle.ts
+++ b/src/useIdle.ts
@@ -46,7 +46,7 @@ const useIdle = (ms: number = oneMinute, initialState: boolean = false, events: 
       }
       off(document, 'visibilitychange', onVisibility);
     };
-  }, events);
+  }, [ms, events]);
 
   return state;
 };
--- a/src/__stories__/useIdle.story.tsx
+++ b/src/__stories__/useIdle.story.tsx
@@ -4,10 +4,12 @@ import {useIdle} from '..';
 import ShowDocs from './util/ShowDocs';
 
 const Demo = () => {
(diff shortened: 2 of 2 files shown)


src/useIdle.ts
```diff
@@ -46,7 +46,7 @@ const useIdle = (ms: number = oneMinute, initialState: boolean = false, events:
       }
       off(document, 'visibilitychange', onVisibility);
     };
-  }, events);
+  }, [ms, events]);
```
streamich: I guess, it should be

```js
[ms, ...events]
```

ManojBahuguna: @streamich
I think it should have been `[events]` in the first place.
Because here `events` is the actual dependency, not items in it.

In fact, if the array updates with less or more events like this:
Before: `['click', 'mousemove']`
After: `['click']`
useEffect won't update and throw a warning saying `The final argument passed to useEffect changed size between renders`.

See this example
https://codesandbox.io/s/n3ow36wql

streamich: Thanks, good to know.

PR #225, opened 2019-04-11
Upgrade babel core and add babel loader for storybook

ankithkonda:
yarn start wasn't working as (i think) babel-loader wasn't being installed as a peer dependency of existing babel 6? so added babel-loader directly

Also, as babel presets were at >v7 and are also part of the @babel scope, thought it'd be best to upgrade babel to v7 using the same scope. 

Now yarn start works as expected and the new test still works.


ankithkonda:
:/ sorry, ignore this

PR #232, opened 2019-04-12
Add TSLint + Prettier and fix lint issues

ankithkonda:
I'll open an issue in case we need to discuss details of this. 
EDIT: Issue opened - #233 :) 

😬 Brace yourselves - a lot of files were changed due to the _autofix_ and some manual changes that I made relating to `no-shadowed-variable`, `no-unused-expression` and `ban-types` (using `()=>void` instead of `Function`)

6UBOB89Y5xZUJ0nIHd-PzSngWXU=


package.json
```diff
@@ -14,6 +14,8 @@
     "start": "yarn storybook",
     "test": "jest",
     "test:watch": "jest --watch",
+    "lint": "node_modules/.bin/tslint src/**/*.{ts,tsx} -t verbose",
+    "lint:fix": "node_modules/.bin/tslint --fix src/**/*.{ts,tsx} -t verbose",
```
streamich: You can just do

```json
{
    "lint": "tslint src/**/*.{ts,tsx} -t verbose",
    "lint:fix": "tslint --fix src/**/*.{ts,tsx} -t verbose",
}
```

src/useClickAway.ts
```diff
@@ -1,11 +1,11 @@
...
+import { off, on } from './util';
 
 const useClickAway = (ref: RefObject<HTMLElement | null>, onClickAway: (event: KeyboardEvent) => void) => {
   useEffect(() => {
-    const handler = (event) => {
-      const {current: el} = ref;
-      el && !el.contains(event.target) && onClickAway(event);
+    const handler = event => {
+      const { current: el } = ref;
+      // el && !el.contains(event.target) && onClickAway(event); // no-unused-expression
```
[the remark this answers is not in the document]
streamich: Probably best to exclude `no-unused-expression` from the TSLint config.

src/useStartTyping.ts
```diff
@@ -19,19 +23,23 @@ const isFocusedElementEditable = () => {
...
   // All other keys.
   return false;
 };
 
 const useStartTyping = (onStartTyping: (event: KeyboardEvent) => void) => {
   useLayoutEffect(() => {
-    const keydown = (event) => {
-      !isFocusedElementEditable() && isTypedCharGood(event) && onStartTyping(event);
+    const keydown = event => {
+      // !isFocusedElementEditable() && isTypedCharGood(event) && onStartTyping(event);  // no-unused-expression
```
streamich: This should not be commented, too.

src/util/createHTMLMediaHook.ts
```diff
@@ -48,43 +50,51 @@ const createHTMLMediaHook = (tag: 'audio' | 'video') => {
...
 
     const wrapEvent = (userEvent, proxyEvent?) => {
-      return (event) => {
+      return event => {
         try {
-          proxyEvent && proxyEvent(event);
+          // proxyEvent && proxyEvent(event);  // no-unused-expression
         } finally {
-          userEvent && userEvent(event);
+          // userEvent && userEvent(event);  // no-unused-expression
```
streamich: These two should not be commented.

tslint.json
```diff
@@ -0,0 +1,37 @@
...
+    "rules": {
+      "interface-name": [true, "never-prefix"],
+      "no-console": false,
+      "max-classes-per-file": false,
+      "no-any": false,
+      "no-empty": false,
+      "prefer-for-of": false,
+      "jsx-no-lambda": [false],
+      "no-empty-interface": [false],
+      "object-literal-sort-keys": false,
```
streamich: Can we please disable `no-empty-interface` and `object-literal-sort-keys`.

ankithkonda: sorry not sure if I understand, they are already set to false

streamich:
I don't have any preferences regarding Prettier (maybe @wardoost has?).

Regarding TSLint: I wouldn't want to set it up strict as to alleviate people, for example sorting object keys by alphabet many developers find annoying; so I would be very mild with TSLint settings.

tslint.json
streamich: Oh, sorry, ignore then.

ankithkonda:
Yup, I agree. I think most of the annoying rules (keys sorting, no empty interface..) should be disabled. I'll push the changes soon, and can wait for @wardoost's feedback to see which others we should disable. :)

streamich:
@ankithkonda I've sent you [a collaborator invite](https://github.com/streamich/react-use/invitations).

ankithkonda:
oh wow! Thank you so much 😃looking forward to working with you all on this 🎉

package.json
```diff
@@ -24,6 +26,11 @@
     "storybook:clean": "rimraf storybook-static",
     "release": "semantic-release"
   },
+  "husky": {
+    "hooks": {
+      "pre-commit": "yarn lint:fix && yarn lint"
```
wardoost: What's the reason for running both lint fix and lint?

wardoost:
With ESLint configs I usually have a warning or error for unused variables like `defaultState` in `useDrop` on [line 25](https://github.com/streamich/react-use/blob/master/src/useDrop.ts#L25). Not sure how that works with tslint as [`no-unused-variable` is deprecated](https://github.com/palantir/tslint/issues/4046).

package.json
ankithkonda: I thought it would fix any lint errors, and finally check to see if there are still errors that it couldn't fix... I just tested it, and it doesn't work :/ woops.. using lint-staged now :)

ankithkonda:
The no-unused-variable deprecation is a pain... but yeah ts guys recommend using these
```
./tsconfig.json -

"noUnusedLocals": true,
"noUnusedParameters": true,
```

Although unused vars not caught by TSLint, I've added in `"ts:type-check": "tsc --noEmit -p tsconfig.json"` script, which I got from here - https://github.com/okonet/lint-staged/issues/138#issuecomment-400618499

It works well with lint-staged now. So the pre-commit checks any ts errors, and then invokes lint-staged which fixes the lint errors. 

Have another go at committing :)

package.json
```diff
@@ -14,6 +14,9 @@
     "start": "yarn storybook",
     "test": "jest",
     "test:watch": "jest --watch",
+    "lint": "tslint src/**/*.{ts,tsx} -t verbose",
```
wardoost: We should add put the glob pattern between parentheses. I have run into the same issue with ESLint before:

> If you don't quote your glob patterns, your shell can intercept them and then we can't guarantee the results. If you quote your glob patterns such that they are not intercepted by the shell, then ESLint can expand the globs in the way you expect. https://github.com/eslint/eslint/issues/9330#issuecomment-330712845

If there is no particular reason to use `verbose` format I would leave the `-t verbose` option out. So this script could look like this:

```
"lint": "tslint 'src/**/*.{ts,tsx}'",
```

package.json
```diff
@@ -14,6 +14,9 @@
...
+    "lint:fix": "tslint --fix src/**/*.{ts,tsx} -t verbose",
```
wardoost: Can be simplified by reusing the `lint` script as yarn passes parameters on

```
"lint:fix": "yarn lint --fix",
```

package.json
```diff
@@ -14,6 +14,9 @@
...
+    "ts:type-check": "tsc --noEmit -p tsconfig.json",
```
wardoost: Do we need to use the `-p tsconfig.json` flag? It seems to work fine without on my machine? We could also rename this script to `lint:types`, not sure if `ts:type-check` is accepted as standard in the TS community?

```
"lint:types": "tsc --noEmit",
```

package.json
```diff
@@ -24,6 +27,11 @@
     "storybook:clean": "rimraf storybook-static",
     "release": "semantic-release"
   },
+  "husky": {
+    "hooks": {
+      "pre-commit": "yarn ts:type-check && lint-staged"
```
wardoost: I would only call `lint-staged` because we only want to do type checking on staged files, not the whole project.
```
"pre-commit": "lint-staged"
```

package.json
```diff
@@ -106,5 +122,11 @@
       "@semantic-release/npm",
       "@semantic-release/git"
     ]
+  },
+  "lint-staged": {
+    "src/**/*.{ts,tsx}": [
+      "yarn lint:fix",
```
wardoost: You normally don't want to use your `lint` scripts as those can lint the whole project with a glob pattern. This could nullify the effect of `lint-staged`. You just have to add the scripts you want to run and `lint-staged` determines which files it will be on. This line would change to

```
"tsc --noEmit && tslint --fix",
```

package.json
streamich: `-t verbose` is useful because it shows which TSLint rules you violate, without it, it just shows line numbers with error messages, but you don't know which rule it is.

image

vs

image

ankithkonda: I agree it's best to keep `-t verbose` for TSLint

package.json
ankithkonda: You're right, there's no need for `-p` in this instance. Also yeah i'm unsure what the standard is for the script name. Happy to go with `lint:types` 👍

package.json
ankithkonda: So I had a go at this, unfortunately, the type checking isn't possible within lint-staged, even when there are unused vars, it still passes. So it seems like TS doesn't support this - https://github.com/okonet/lint-staged/issues/468#issuecomment-418827220 

When I run the type checking outside lint-staged it works well. It takes about 5s on my machine for it to check the whole project, I think this is okay.

package.json
ankithkonda: added glob between parentheses 👍

package.json
ankithkonda: done! 👍

package.json
ankithkonda: renamed the script and removed `-p tsconfig.json`

package.json
ankithkonda: see above 👆 :)

package.json
wardoost: Should have read that link you originally posted a bit more. This looks like the best solution at the moment 👍

wardoost:
This looks really great, thanks @ankithkonda 👍

ankithkonda:
No worries 👍 Shall I go ahead and merge?

PR #241, opened 2019-04-16
useResizeObserver + useForceUpdate

samdenty:
I've created a `useResizeObserver` hook, that replaces the existing `useSize` hook.

I've left the useSize hook, so you can decide whether to deprecate it / remove it altogether

It supports observing the dimensions & x/y coords, selectively updating the component depending on which properties you access


samdenty:
You should really add a `.prettierrc` and add `pretty-quick` to pre-commit

ankithkonda:
tslint and prettier have already been added see - #232 :)

if you use vscode, you can use the inbuilt editor code actions to fix on save... here's what i use - 
```
{
  "editor.codeActionsOnSave": {
    //"source.fixAll.tslint": true
  }
}
```

however, #232 does include the ability to lint and fix for precommit

ankithkonda:
I'd recommend adding tests for your new hooks 👍

samdenty:
@ankithkonda I was going to, but saw most of the hooks weren't tested.

samdenty:
> tslint and prettier have already been added see - #232 :)

There's no (`.prettierrc`)[https://github.com/streamich/react-use/blob/master/.prettierrc] in the repo, and the prettier settings seem to be non-default

streamich:
Prettier config is [here](https://github.com/streamich/react-use/blob/master/tslint.json#L26-L34), as I understand it.

docs/useForceUpdate.md
```diff
@@ -0,0 +1,20 @@
+# `useForceUpdate`
```
streamich: How is this hook better/different from [`useUpdate`](https://github.com/streamich/react-use/blob/master/docs/useUpdate.md)?

streamich:
What is the reasoning for using `Proxy` + `useForceUpdate` instead of saving the dimensions into a local state using `useState`?

streamich:
@samdenty I've simplified the PR and created a new one [here](https://github.com/streamich/react-use/pull/266), please let me know if removal of `Proxy` will have some negative consequences.

samdenty:
@streamich The reason I used a proxy is because it prevents unnecessary renders.

```ts
// The component will only update on width change, if the height changes (or any other property) it doesn't matter
const { width } = useResizeObserver()

// Will only update when x or y change
const { x, y } = useResizeObserver()
```

With your useResizeObserver hook it only supports `width` + `height`, and the component will will have unnecessary renders:


- The `top` / `bottom` / `x` / `y` property of an element changes
- When `height` or `width` changes, regardless of whether are using both in the component

Issue #164, opened 2019-03-24
useReducer middleware


wardoost:
Extend the standard `useReducer` with a dispatcher that can handle "thunks" similar to [`redux-thunk`](https://github.com/reduxjs/redux-thunk). It could look something like this:
```js
export const useThunkReducer = (reducer, initialState) => {
  const [state, dispatch] = useReducer(reducer, initialState);

  const thunkDispatch = action => {
    if (typeof action === "function") {
      return action(dispatch, state);
    }

    return dispatch(action);
  };

  return [state, thunkDispatch];
}
```

Something similar was described in this article: https://medium.com/yld-engineering-blog/rolling-your-own-redux-with-react-hooks-and-context-bbeea18b1253

streamich:
An interesting hook would be that allows to add any Redux middleware.

```js
const useMyReducer = applyMiddleware(useReducer, [logging, thunks, ...]);
```

wardoost:
I have created a [`createReducer` factory function](https://github.com/streamich/react-use/blob/createReducer/src/createReducer.ts) the that takes middleware like Redux's [`applyMiddleware`](https://redux.js.org/api/applymiddleware#applymiddlewaremiddleware) and returns a reducer that can be used like React's `useReducer`. While it looks handy to be able to use redux middleware, I am not sure if it's useful enough to add to this library.

Also, there are already a few repositories that have been doing Redux things with hooks ([`redux-react-hook`](https://github.com/facebookincubator/redux-react-hook), [`redux-hooks`](https://github.com/epeli/redux-hooks)) but experienced issues and `react-redux` is working on a hooks API: https://github.com/reduxjs/react-redux/issues/1179.

wardoost:
I noticed that `getState` always  returns the current `state` which is problematic for redux middleware. With the current storybook example you can see the `prev state` and `next state` are the same which is incorrect behaviour:

<img width="307" alt="Screen Shot 2019-04-20 at 12 11 22 pm" src="https://user-images.githubusercontent.com/5230018/56450178-1369e880-6367-11e9-9113-60de0f1dbfdb.png">

In order to fix this we need a `getState` function that always returns the correct state. As far as I know this is not possible with `useReducer`?

PR #244, opened 2019-04-17
refacotr: memorize getState and applyMiddleware

ariesjia:


ariesjia:
#164

ariesjia:
fixed `getState` always return  the initial state issue , reducer will always return a new object , so we can memoize the `getState` method as state changes.

now the storybook demo works as previously

wardoost:
While I see memoizing `getState`/`middlewareAPI` is needed when using this hook in combination with other hooks this branch also suffers from the issue mentioned in https://github.com/streamich/react-use/issues/164#issuecomment-485051613.

PR #255, opened 2019-04-22
chore: use useState to mock reducer to make getState always return current state

ariesjia:


wardoost:
Nice take on this but there still seems to be an issue with async action:

Screen Shot 2019-05-02 at 7 08 24 pm

It's also better to add a little description of what you're trying to do, makes it a bit easier to review ✌️

ariesjia:
the async action issue seems no way to solve it.
getState method only return current render state, can't get the next render state value.

PR #266, opened 2019-04-23
Use resize observer 2

streamich:


src/__stories__/useResizeObserver.story.tsx
```diff
@@ -0,0 +1,23 @@
+import { storiesOf } from '@storybook/react';
+import * as React from 'react';
+import { useRef } from 'react';
+import { useResizeObserver } from '..';
+import ShowDocs from './util/ShowDocs';
+
+const Demo = () => {
+  const ref = useRef(null);
+  const { width, height } = useResizeObserver(ref);
```
wardoost: `useRef` can actually be placed in the hook and return a tuple with state and ref object so it can be used like this:

```js
const [{ width, height }, ref] = useResizeObserver();
```

We might want to change this in hooks that are using this pattern too.

PR #268, opened 2019-04-23
Add ability to pass position options to useGeolocation

denysdovhan:
This PR adds ability to pass [position options](https://developer.mozilla.org/docs/Web/API/PositionOptions) for `useGeolocation` hook.

### DIFF
--- a/src/useGeolocation.ts
+++ b/src/useGeolocation.ts
@@ -13,7 +13,7 @@ export interface GeoLocationSensorState {
   error?: Error | PositionError;
 }
 
-const useGeolocation = (): GeoLocationSensorState => {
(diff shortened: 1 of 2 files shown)


src/useGeolocation.ts
```diff
@@ -13,7 +13,7 @@ export interface GeoLocationSensorState {
   error?: Error | PositionError;
 }
 
-const useGeolocation = (): GeoLocationSensorState => {
+const useGeolocation = (options: PositionOptions): GeoLocationSensorState => {
```
streamich: Can we make the options optional?

```
options?: Posi...
```

PR #294, opened 2019-05-08
usetoggle no param casting

### LINKED ISSUE #278: useToggle shouldn't try to cast its param to boolean
I think trying to convert the `nextValue` param to boolean in `useToggle` is wrong. Most of the times we want to pass the `toggle` function directly to a component as a prop, but because of https://github.com/streamich/react-use/blob/master/src/useToggle.ts#L16 we can't because it will cast, for example, an event to `true` (imagine `onClick={toggle}`), which does not give the desired behavior (sets `value` to `true` always)

CodeSandbox repro: https://codesandbox.io/s/km341w2jpo

We now have to wrap our toggle in another function if we effectively want it to toggle.

Just check if `typeof nextValue === 'boolean'` and set that, else negate previous value.

wardoost:
Ignore non-boolean values parameters  of `useToggle` updater function and toggle state as if there was no parameter. This allows you to use the update action like `onClick={toggle}`.

Fixes #278

### DIFF
--- a/docs/useToggle.md
+++ b/docs/useToggle.md
@@ -4,7 +4,6 @@ React state hook that tracks value of a boolean.
 
 `useBoolean` is an alias for `useToggle`.
 
-
 ## Usage
 
 ```jsx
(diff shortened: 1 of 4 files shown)


wardoost:
@ankithkonda I have commented out two tests as I'm getting a state update after unmount React error. Not sure what I'm doing wrong.

PR #315, opened 2019-05-24
Create reducer

streamich:


streamich:
Seems like these changes fail to build.

ariesjia:
It's just typing error ,  we can replace the 

```js
let middlewareDispatch = () => {
    throw new Error(
      'Dispatching while constructing your middleware is not allowed. ' +
        'Other middleware would not be applied to this dispatch.'
    );
  };
```
with 
```js
let middlewareDispatch = (_ = {}) => {
    throw new Error(
      'Dispatching while constructing your middleware is not allowed. ' +
        'Other middleware would not be applied to this dispatch.'
    );
};
```
or remove the warning
```js
let middlewareDispatch;
```

PR #321, opened 2019-05-28
add tests, and fix the initial state issue

### LINKED ISSUE #314: useAsync: loading is false on mount
Due to a change in v8.1, the behavior and typings of `useAsync` are no longer accurate.

https://github.com/streamich/react-use/commit/4b998d6e0834e29f366574d668dc361c50be7ef3#diff-8e6e0808e6745d8058a2f9cbd476172dR22

Before that commit, there was a guarantee that if `loading` was false, then either `error` or `value` was set on the state.  Since that commit, `loading` is now false at mount, but `error` and `value` are both undefined.

That means that the following code errors on mount, even though it passes type checks:

```typescript
function App() {
  const state = useAsync(getUsers, []);

  if (state.loading) {
    return <>...</>;
  }

  if (state.error) {
    return <>Something went wrong!</>;
  }

  return (
    <ul>
      {state.value.map((user, i) => (
        <li key={i}>{user.name}</li>
      ))}
    </ul>
  );
}
```

Here's an example: https://uegk4.codesandbox.io/

NullVoxPopuli:
resolves #314 

Also, improved type inference:
image


babel.config.js
```diff
@@ -1,3 +1,14 @@
 module.exports = {
-    presets: ['@babel/preset-env', '@babel/preset-react', "@babel/preset-typescript"],
-};
\ No newline at end of file
+  presets: [
+    [
+      "@babel/preset-env",
+      {
+        targets: {
+          node: "current"
```
NullVoxPopuli: allow the regenerator runtime to be used

package.json
```diff
@@ -128,5 +128,9 @@
       "tslint --fix -t verbose",
       "git add"
     ]
+  },
+  "volta": {
```
NullVoxPopuli: pinned versions of tools for react-use development purposes
this is using volta-cli

src/__tests__/useAsync.test.tsx
```diff
@@ -0,0 +1,131 @@
+import { useCallback } from 'react';
+import { cleanup, renderHook } from 'react-hooks-testing-library';
+import useAsync from '../useAsync';
+
+afterEach(cleanup);
+
+// NOTE: these tests cause console errors.
```
NullVoxPopuli: idk what to do about this. :man_shrugging: 
never used react-hooks-testing-library before, and I'm generally not a fan of _not_ testing UI -targeting code outside the browser :-\

src/useAsync.ts
```diff
@@ -18,14 +18,18 @@ export type AsyncState<T> =
       value: T;
     };
 
-const useAsync = <T>(fn: () => Promise<T>, deps: DependencyList = []) => {
-  const [state, callback] = useAsyncFn(fn, deps);
+export default function useAsync<
+  Result = any,
```
NullVoxPopuli: this level of splitting up of the types allows everything to be type-aware, even the args for the callback.
previously, everything was any all the things

tslint.json
```diff
@@ -1,18 +1,17 @@
...
+    "tslint-react",
+    "tslint-eslint-rules",
+    "tslint-config-prettier"
+  ],
+  "linterOptions": {
+    "exclude": ["node_modules/**"]
+  },
+  "jsRules": {},
+  "rules": {
+    "ban-types": false,
```
NullVoxPopuli: `Function` was previously a disallowed type and `Function` is needed in order to declare args as invokeable.

as a side note, this project should ditch tslint in favor of eslint, as tslint has been abandoned.

src/useAsyncFn.ts
```diff
@@ -18,17 +18,23 @@ export type AsyncState<T> =
...
-    loading: false,
-  });
+export default function useAsyncFn<
+  Result = any,
+  Args extends any[] = any[],
+  Fn extends Function = (...args: Args | []) => Promise<Result>
+>(
+  fn: Fn,
+  deps: DependencyList = [],
+  initialState: AsyncState<Result> = { loading: false }
```
NullVoxPopuli: this initial state here resolves the regression introduced by https://github.com/streamich/react-use/commit/4b998d6e0834e29f366574d668dc361c50be7ef3#diff-8e6e0808e6745d8058a2f9cbd476172dR22

src/useAsync.ts
```diff
@@ -18,14 +18,18 @@ export type AsyncState<T> =
...
+  Args extends any[] = any[],
+  Fn extends Function = (...args: Args | []) => Promise<Result>
+>(fn: Fn, deps: DependencyList = []) {
+  const [state, callback] = useAsyncFn<Result, Args, Fn>(fn, deps, {
+    loading: true,
```
NullVoxPopuli: this initial state here resolves the regression introduced by https://github.com/streamich/react-use/commit/4b998d6e0834e29f366574d668dc361c50be7ef3#diff-8e6e0808e6745d8058a2f9cbd476172dR22

src/__tests__/useAsync.test.tsx
streamich: I think it is pretty standard to test React components or hooks using Jest.

NullVoxPopuli: I'm challenging that standard ;)

src/useAsyncFn.ts
```diff
@@ -18,17 +18,23 @@ export type AsyncState<T> =
...
-const useAsyncFn = <T>(fn: (...args: any[]) => Promise<T>, deps: DependencyList = []): [AsyncState<T>, () => void] => {
-  const [state, set] = useState<AsyncState<T>>({
-    loading: false,
-  });
+export default function useAsyncFn<
+  Result = any,
+  Args extends any[] = any[],
+  Fn extends Function = (...args: Args | []) => Promise<Result>
+>(
+  fn: Fn,
```
streamich: Same here, maybe we can just do

```ts
fn: (...args: Args | []) => Promise<Result>
```

and get rid of `Fn`?

streamich: Alternatively, a fancy way would be to just specify `Fn` and get rid of `Args` and `Result` but instead extract them where needed. Only thing then `Args` are not often needed and users will always have to type `Promise<...>` although it is already implied.

src/useAsyncFn.ts
```diff
@@ -18,17 +18,23 @@ export type AsyncState<T> =
...
+  const [state, set] = useState<AsyncState<Result>>(initialState);
 
   const mounted = useRefMounted();
 
-  const callback = useCallback((...args) => {
+  const callback = useCallback((...args: Args | []) => {
     set({ loading: true });
 
-    fn(...args).then(
+    return fn(...args).then(
```
streamich: nice

What is you use case for returning value? Also I think, if you want to return it here, you also need to return it from the `.then(value => { return value; })`.

tslint.json
streamich: agreed

src/__tests__/useAsync.test.tsx
NullVoxPopuli: it'd be a pretty big change, but https://www.bigtestjs.io/ (a layer on top of mocha)
It's more for UI testing, but mocha also does unit testing.

The main thing is that we'd be able to set up real react scenarios and test interaction with real components in a real browser in a real dom. 

(users don't use fake dom, is my main point here)

src/useAsync.ts
```diff
@@ -18,14 +18,18 @@ export type AsyncState<T> =
       value: T;
     };
 
-const useAsync = <T>(fn: () => Promise<T>, deps: DependencyList = []) => {
-  const [state, callback] = useAsyncFn(fn, deps);
+export default function useAsync<
+  Result = any,
+  Args extends any[] = any[],
+  Fn extends Function = (...args: Args | []) => Promise<Result>
```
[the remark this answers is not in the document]
NullVoxPopuli: I actually don't know. hm

src/useAsyncFn.ts
NullVoxPopuli: can't get rid of Args and Result, or otherwise the calling signature from the returned callback won't be typed or callable, I think. I forgot the specific issue I ran in to

NullVoxPopuli: Args are needed for an app I'm working on, as I'm doing a lot of parameter passing with useAsyncFn

src/useAsyncFn.ts
NullVoxPopuli: mostly so we can await the callback

```ts
const [state, callback] = ...

// ...

//ideally,
const myReturnValue = await callback
// probably need to return value and return error in each of the promise callbacks though. hmm.

need more tests
```

NullVoxPopuli:
>Thanks! This is very useful, especially the tests for useAsync* functions as previously these hooks were changed a lot and now we have some spec for them.

<3

> I am just curious on your thoughts about Fn type and returning value from .then(...), otherwise looks good for merging.

Yeah, I'll need to explore these a bit, don't know if I'll get to it today though

src/__tests__/useAsync.test.tsx
NullVoxPopuli: alternatively, if the react team had more ergonomic testing tools for hooks that considered the async / effect case, I think that would still get us a long way

NullVoxPopuli: my main thing with the current react-hooks-testing-library are the console errors, and how the API for getting the next value is kinda awkward -- requires a `.rerender(..props)` and `waitForNextUpdate` :-\

src/useAsyncFn.ts
streamich: I mean either keep (`Args` + `Restult`) or just `Fn`.

NullVoxPopuli: ah! turns out that my reason for adding Fn was removed when I return value from the then callback. noice

NullVoxPopuli: adding more tests now. woosh

NullVoxPopuli:
ok, I removed the Fn type, and added a test (and supporting code) to assert that when you await the callback, you'll get the return value of `fn`

PR #327, opened 2019-05-30
Fixed useSetState definition

bob76828:
Correct the definition that the setState updater function must to have the prevState parameter.


bob76828:
duplicate #323

Issue #370, opened 2019-06-08
useThrottle has missing dependencies


steida:
React Hook useEffect has missing dependencies: 'hasNextValue', 'ms', and 'nextValue'. Either include them or remove the dependency array.eslint(react-hooks/exhaustive-deps)

steida:
I see the project is using tslint instead of eslint. Not using react-hooks/exhaustive-deps is dangerous. I will send PR.

steida:
Ups.

```
/Users/steida/dev/react-use/src/createMemo.ts
  4:45  error  React Hook useMemo was passed a dependency list that is not an array literal. This means we can't statically verify whether you've passed the correct dependencies  react-hooks/exhaustive-deps
  4:45  error  React Hook useMemo has a missing dependency: 'args'. Either include it or remove the dependency array                                                               react-hooks/exhaustive-deps

/Users/steida/dev/react-use/src/createReducer.ts
  27:6  error  React Hook useMemo was passed a dependency list that is not an array literal. This means we can't statically verify whether you've passed the correct dependencies  react-hooks/exhaustive-deps

/Users/steida/dev/react-use/src/useAsyncFn.ts
  54:6  error  React Hook useCallback was passed a dependency list that is not an array literal. This means we can't statically verify whether you've passed the correct dependencies                                                                 react-hooks/exhaustive-deps
  54:6  error  React Hook useCallback has missing dependencies: 'fn' and 'mounted'. Either include them or remove the dependency array. If 'fn' changes too often, find the parent component that defines it and wrap that definition in useCallback  react-hooks/exhaustive-deps

/Users/steida/dev/react-use/src/useAsyncRetry.ts
  23:6  error  React Hook useCallback has an unnecessary dependency: 'attempt'. Either exclude it or remove the dependency array                                          react-hooks/exhaustive-deps
  23:7  error  React Hook useCallback has a spread element in its dependency array. This means we can't statically verify whether you've passed the correct dependencies  react-hooks/exhaustive-deps

/Users/steida/dev/react-use/src/useBattery.ts
  37:19  error  Assignments to the 'battery' variable from inside React Hook useEffect will be lost after each render. To preserve the value over time, store it in a useRef Hook and keep the mutable value in the '.current' property. Otherwise, you can move this variable directly inside useEffect  react-hooks/exhaustive-deps
[...]

PR #371, opened 2019-06-09
Add eslint with react-hooks plugin

steida:
react-hooks plugin with the exhaustive check is the must.

Just check how many weird things and errors the code base contains: https://github.com/streamich/react-use/issues/370


wardoost:
Great 👏 We should probably remove tslint at the same time?

steida:
@wardoost Yes, it should be removed. The teams behind tslint and eslint for TypeScript are collaborating.

streamich:
Some thoughts on linter situation (in general, not related to this PR):

- [ ] Remove TSLint.
- [ ] Enable ESLint with similar rules to TSLint, but more relaxed (as current TSLint rules seem too strict).
- [ ] Enable React hook rules in ESLint.
- [ ] Remove Prettier rules from TSLint/ESLint.
- [ ] Run Prettier separately `prettier ...` on pre-commit hook. (Or probably run `pretty-quick` instead.)

steida:
Check https://github.com/este/este/blob/master/.eslintrc.js for inspiration.

wardoost:
@streamich Wouldn't it be better to integrate Prettier in ESLint to avoid conflicting rules/autofixes? Are there any specific TSLint rules that you think of now that are too strict?

streamich:
@wardoost 

> ... any specific TSLint rules that you think of now that are too strict?

There is one or more rules that require you to sort stuff in alphabetical order, don't remember if it was object keys or imports. Also, there was something else [don't remember now] ...

Why I would like to see Prettier outside of ESLint is so that commit or push never fails because of so code style problem. Prettier can be run any time on the whole project to fix all style "errors", I don't see necessity to fail a commit because of it, or for syntax highlighter to highlight all Prettier things as "errors".

> Wouldn't it be better to integrate Prettier in ESLint to avoid conflicting rules/autofixes?

- Prettier [can be run any time](https://github.com/streamich/memfs/blob/master/package.json#L18), to re-format the whole project.
- Also it can be run automatically run on every commit, so you don't need to ever reformat the whole project.
- For TSLint there is `tslint-config-prettier` plugin that disables all stylistic rules so that TSLint does not conflict with Prettier, ESLint should have something similar.

wardoost:
Sounds perf 👌I also like the suggested config from @steida as a starting point, minus prettier and relay plugins

ayush987goyal:
Hi @streamich ,

I have done some local `eslint` configuration with the default `eslint-config-react-app` plugin and I am seeing these many errors/warnings:
```
✖ 96 problems (18 errors, 78 warnings)
  0 errors and 52 warnings potentially fixable with the `--fix` option.
```
I was thinking of adding `eslint-disable` to all the affected file and merging that change. Then we have to fix the files one by one and create respective CRs.
I am down to do that but would require contributions along the way.
This would be a major version change since this might potentially break existing functionalities.
So, how should we go about this change?

streamich:
@ayush987goyal I think `eslint-config-react-app` is a good option, unless there are other propositions.

> I was thinking of adding eslint-disable to all the affected file and merging that change. Then we have to fix the files one by one and create respective CRs.

Sounds good to me.

ayush987goyal:
@streamich Here is #948 to start this.

PR #374, opened 2019-06-11
Add useScrolling hook

james2406:
This hook keeps track of whether the user is scrolling or not. It can be used like this:

```javascript
const scrolling = useScrolling(scrollerRef);
```

Or can be used with the `usePrevious` hook to detect when a scroll starts or ends:

```javascript
const scrolling = useScrolling(scrollerRef);
const prevScrolling = usePrevious(scrolling);

useEffect(
  () => {
    if (scrolling && !prevScrolling) {
      // scroll started
    } else if (!scrolling && prevScrolling) {
      // scroll ended
    }
  },
  [scrolling],
);
```

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -52,6 +52,7 @@ import usePromise from './usePromise';
 import useRaf from './useRaf';
 import useRefMounted from './useRefMounted';
 import useScroll from './useScroll';
+import useScrolling from './useScrolling';
 import useSessionStorage from './useSessionStorage';
 import useSetState from './useSetState';
 import useSize from './useSize';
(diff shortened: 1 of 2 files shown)


streamich:
Would this work, too?

```js
useEffect(
  () => {
    if (scrolling) {
      // scroll started
      return () => {
        // scroll ended
      };
    }
  },
  [scrolling],
);
```

james2406:
> Would this work, too?
> 
> ```js
> useEffect(
>   () => {
>     if (scrolling) {
>       // scroll started
>       return () => {
>         // scroll ended
>       };
>     }
>   },
>   [scrolling],
> );
> ```

Yeah, it would! Much cleaner

Issue #172, opened 2019-03-26
useAsync should keep last value when loading again.


xitianfz:
Currently useAsync would set its `state.value` to `undefined` when initiate another loading. This is sometimes inconvenient when used in codes as below. It makes the Table empty during the loading instead of showing old data.

```jsx
const dataLoader = useAsync(...)

return (
  <Table data={dataLoader.value} loading={dataLoader.loading} />
)
```

Is it possible to keep last value when loading again?

wardoost:
I agree, the other case where you don't want to display old data while loading could still be done like this

```js
const dataLoader = useAsync(...)

return (
  <Table
    data={dataLoader.loading ? undefined : dataLoader.value}
    loading={dataLoader.loading}
  />
)
```

Or you could do the loading check in your `Table` component

wardoost:
We could use `useSetState` instead of `React.useState` to manage `useAsync`'s internal state so that it always returns a `value`, `loading` and `error` keys. Or is there a reason for not returning a state of the same shape every time?

oliverviljamaa:
Agreed, we ended up creating a custom hook for this behaviour.

streamich:
Closing as this seems old issue. Feel free to re-open.

PR #381, opened 2019-06-13
feat: preserve the previous state of useAsync while loading

blackLearning:
make the initial state works as expected, like the usual react states


xobotyi:
@streamich now it's mergable.
What do you think? First time i've used this hook i tried to pass the state with custom fields and been wondered when they didnt appear in state.
Thus, in my opinion, it should be preserved here(mostly because users can't preserve it on demand due to inablitity to controll inner useState);

shrynx:
@streamich 
Hi 👋
do you  think this can be merged in ?
Has been creating issues and temporarily have to create a custom hook for this
Also has been separately reported in issue as well
https://github.com/streamich/react-use/issues/172

streamich:
It seems like a breaking change.

1. Shall we keep the previous state in a separate property, say `prev`?

```js
const { prev } = useAsync();
```

2. Should this functionality be opt-in?

```js
useAsync(fn, { preserve: true })
```

I've mentioned this change in a [related discussion](https://github.com/streamich/react-use/pull/581#issuecomment-552252508).

xobotyi:
IMO it should not be neither optional nor stored in other field.
As an example useState when you passes function there does not flush current state. I know that its not 100% suitable example but hope you got the sense.

mario-jerkovic:
I am really looking forward to merging this as it's really unexpected that the loading or error states flushes current value state.

PR #394, opened 2019-06-14
feat: improve useClickAway() hook

jeetiss:


src/useClickAway.ts
```diff
@@ -8,10 +8,14 @@ const useClickAway = (
   onClickAway: (event: KeyboardEvent) => void,
   events: string[] = defaultEvents
 ) => {
+  const savedCallback = useRef(onClickAway);
```
streamich: Instead of saving the callback inside the `useClickAway` hook, isn't it better if consumer do it manually?

```js
const onClickAway = useCallback(() => { /* ... */ });
useClickAway(ref, onClickAway);
```

src/useClickAway.ts
```diff
@@ -21,7 +25,7 @@ const useClickAway = (
         off(document, eventName, handler);
       }
     };
-  });
+  }, [events, ref]);
```
streamich: `events` is an array, I though array cannot be a *dep*, maybe it should be like this:

```js
[...events, ref]
```

src/useClickAway.ts
jeetiss: I don't want to write extra code on every `useClickAway` usage
and this approach is less effective

src/useClickAway.ts
jeetiss: profs?)

> React Hook useEffect has a spread element in its dependency array. This means we can't statically verify whether you've passed the correct dependencies. (react-hooks/exhaustive-deps)

jeetiss: you're right

https://codesandbox.io/s/deps-array-kj8ov

jeetiss: sorry, I'm trying to figure it out myself.

And you approach is totally wrong, with spread operator in `deps` react writes this warnings

```

Warning: The final argument passed to useEffect changed size between renders. The order and size of this array must remain constant.

Previous: [1, 2, 3, 1, 1, 1]
Incoming: [1, 2, 3]
    in App
```

streamich: Could `useEffect` be changed to [`useDeepCompareEffect`](https://github.com/streamich/react-use/blob/master/docs/useDeepCompareEffect.md)?

jeetiss: Why? `useEffect` works fine!
See codesandbox example

streamich: I guess you are right. I was thinking if `events` is provided inline, it will be a new array every re-render.

```js
useClickAway(ref, fn, ['mousedown', 'touchstart'])
```

src/useClickAway.ts
morijenab: may I ask what is the purpos of using the useRef here ?
why not simply use the callback ?

PR #411, opened 2019-06-24
fix: createReducer

Andrey-Bazhanov:
- make docs for `createReducer ` identical to story
- fix `dispatch` creation on every render see [demo](https://codesandbox.io/s/react-use-createreducer-bug-o0po8)

### DIFF
--- a/docs/createReducer.md
+++ b/docs/createReducer.md
@@ -7,17 +7,11 @@ Factory for reducer hooks with custom middleware with an identical API as [React
 An example with [`redux-thunk`](https://github.com/reduxjs/redux-thunk) and [`redux-logger`](https://github.com/LogRocket/redux-logger).
 
 ```jsx
-import React from 'react';
 import { createReducer } from 'react-use';
 import logger from 'redux-logger';
 import thunk from 'redux-thunk';
 
 const useThunkReducer = createReducer(thunk, logger);
-
-// React useReducer lazy initialization example: https://reactjs.org/docs/hooks-reference.html#lazy-initialization
-function init(initialCount) {
-  return { count: initialCount };
-}
 
 function reducer(state, action) {
   switch (action.type) {
@@ -26,7 +20,7 @@ function reducer(state, action) {
     case 'decrement':
       return { count: state.count - 1 };
     case 'reset':
-      return init(action.payload);
+      return { count: action.payload };
     default:
       throw new Error();
   }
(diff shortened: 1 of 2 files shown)


src/createReducer.ts
```diff
@@ -8,29 +8,40 @@ function composeMiddleware(chain) {
...
+    const dispatch = useCallback(
+      action => {
+        ref.current = reducer(ref.current, action);
+        setState(ref.current);
+        return action;
+      },
+      [reducer]
     );
+
+    useMemo(() => {
```
wardoost: These changes look great 👌

Can you explain why you use `useMemo` instead `useEffect` here?

docs/createReducer.md
```diff
@@ -20,31 +20,38 @@ function reducer(state, action) {
...
       setTimeout(() => {
-        dispatch({ type: 'reset', payload: 1 });
+        dispatch({ type: 'reset', payload: initialCount });
       }, 1000);
     };
-  }, []);
+  }, [initialCount]);
 
-  const [count, dispatch] = useThunkReducer(reducer, 1);
+  const [state, dispatch] = useThunkReducer(reducer, initialCount, init);
```
wardoost: Can remove `init`

docs/createReducer.md
```diff
@@ -20,31 +20,38 @@ function reducer(state, action) {
...
 
   return (
     <div>
-      <p>count: {count}</p>
+      <pre>{JSON.stringify(state, null, 2)}</pre>
```
wardoost: Could we use `<p>count: {state.count}</p>` here, `JSON.stringify` in a code example might be a bit confusing?

src/createReducer.ts
wardoost: Could we use `useIsomorphicLayoutEffect` for this?

Andrey-Bazhanov: @wardoost I simplified createReducer - now we no need any useMemo or useEffect (all useReducer functionality still available)

src/createReducer.ts
```diff
@@ -8,29 +8,31 @@ function composeMiddleware(chain) {
...
+
+  return (reducer, initialState, initializer = value => value) => {
+    const ref = useRef(initializer(initialState));
+    const [, setState] = useState(ref.current);
+
+    const dispatch = useCallback(action => {
+      ref.current = reducer(ref.current, action);
+      setState(ref.current);
+      return action;
+    }, []);
```
wardoost: I'm not sure if `reducer` should be added to the dependency array? As far as I know it's not common to swap out a reducer with`useReducer`?

Andrey-Bazhanov: Yeah you are right. We can swap [reducers](https://codesandbox.io/s/nostalgic-proskuriakova-q85mr) with `useReducer`. So we need add `reducer` to the dependency array. And add

```
  useMemo(() => {
    dispatchRef.current = composedMiddleware(
      {
        getState: () => ref.current,
        dispatch: (...args) => dispatchRef.current(...args),
      },
      dispatch
    );
  }, [dispatch]);
```

because with `useEffect` we have double dispatch initialization [screencast](https://take.ms/l3N88)

What do you think?

wardoost: Would [useUpdateEffect](https://github.com/streamich/react-use/blob/master/docs/useUpdateEffect.md) fix the double dispatch initialization?

Andrey-Bazhanov: With `useUpdateEffect` works pretty nice.

Andrey-Bazhanov:
@wardoost I think pr is ready. Can you confirm?

PR #415, opened 2019-06-25
Feat: add voice option in use speech

gelove:


streamich:
@gelove I believe you didn't want to close this PR?

PR #421, opened 2019-06-26
feat: add usePermission

### LINKED ISSUE #428: usePermission
This is a proposal to query permission status.
## **Usage**
```
import { usePermission } from 'react-use';

type PermissionName = "geolocation" | "notifications" | "push" | "midi" | "camera" | "microphone" | "speaker" | "device-info" | "background-sync" | "bluetooth" | "persistent-storage" | "ambient-light-sensor" | "accelerometer" | "gyroscope" | "magnetometer" | "clipboard";

const Demo = () => {
  const { status } = usePermission({name: 'microphone'});

  return (
    <div>
        {status === 'denied' && <span> Unable to access microphone，has no permission.</span>}
         {status === 'granted' && <span> Have access to microphone. </span>}
    </div>
  );
}
```

gelove:
Closes #428

### DIFF
--- a/src/usePermission.ts
+++ b/src/usePermission.ts
@@ -1,0 +1,46 @@
+import { useEffect, useState } from 'react';
+import { off, on } from './util';
+
+type PermissionDesc =
+  | PermissionDescriptor
+  | DevicePermissionDescriptor
+  | MidiPermissionDescriptor
+  | PushPermissionDescriptor;
+
+const noop = () => {};
+
+const usePermission = (permissionDesc: PermissionDesc): string => {
+  const [state, setState] = useState('');
+  let mounted = true;
+  let permissionStatus: PermissionStatus | null = null;
+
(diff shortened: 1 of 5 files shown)


src/usePermission.ts
```diff
@@ -0,0 +1,46 @@
...
+
+type PermissionDesc =
+  | PermissionDescriptor
+  | DevicePermissionDescriptor
+  | MidiPermissionDescriptor
+  | PushPermissionDescriptor;
+
+const noop = () => {};
+
+const usePermission = (permissionDesc: PermissionDesc): string => {
```
streamich: Instead of `string` return type, maybe we can make it more specific?

```ts
'' | 'granted' | 'denied' | 'prompt'
```

gelove: Thanks for your suggestion.

PR #440, opened 2019-07-04
Update useAsync.ts

sunxing102005:
add arguments to async function

### DIFF
--- a/src/useAsync.ts
+++ b/src/useAsync.ts
@@ -5,14 +5,15 @@ export { AsyncState, AsyncFn } from './useAsyncFn';
 
 export default function useAsync<Result = any, Args extends any[] = any[]>(
   fn: (...args: Args | []) => Promise<Result>,
-  deps: DependencyList = []
+  deps: DependencyList = [],
+  ...callbackArgs: Args | []
 ) {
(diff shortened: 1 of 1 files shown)


wardoost:
@sunxing102005 Could you give an example where this would be necessary?

sunxing102005:
if the async function has parameters ,the old useAsync can not add them.
```
const fn = (param)=>{
//it is an async function to fetch data
 return axios(param);
}
//we use the old useAsync like this
const state = useAsync(fn);
// but no way to set fn its params

//after my edit,we can use it like this
const state = useAsync(fn,[],param);

```

wardoost:
Wouldn't you write that like this?
```js
const state = useAsync(() => {
  return axios(param);
}, [param]);
```
This is more in line with the API design of React's standard hooks. Besides that the `param` variable is probably coming from the props or state of the component it is used in. So if these parameters change your async function needs to be triggered again so you want to add `param` to the hook's dependencies.

sunxing102005:
yes you are right, that 's a better solution, and  thank you for the interpretation！

wardoost:
Np 👌

PR #453, opened 2019-07-11
Add useInterval hook and fix useDebounce hook's bug

lee1994522:
Add useInterval hook


streamich:
Thanks! I will merge it here: https://github.com/streamich/react-use/pull/470

PR #454, opened 2019-07-11
useMedia - initialize state with call to media query

mattleonowicz:
Initializing state with result from call to window.matchMedia avoids possible initial flicker of returned value, when media query matched initially


mattleonowicz:
Thanks!

PR #458, opened 2019-07-12
Add dependency "@types/react-wait"

### LINKED ISSUE #306: useWait typings
Hello! I think there are some missing typings for useWait, got the following error:

```
  ERROR in [ProjectDirectory]/node_modules/react-use/lib/useWait.d.ts(1,25):
  TS7016: Could not find a declaration file for module 'react-wait'. '[ProjectDirectory]/node_modules/react-wait/dist/react-wait.js' implicitly has an 'any' type.
    Try `npm install @types/react-wait` if it exists or add a new declaration (.d.ts) file containing `declare module 'react-wait';`
```

It should be served by `react-use`, without adding `@types/react-wait`. Dependency?

elliottsj:
Fixes #306

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
   },
   "homepage": "https://github.com/streamich/react-use#readme",
   "dependencies": {
+    "@types/react-wait": "^0.3.0",
     "copy-to-clipboard": "^3.1.0",
     "nano-css": "^5.1.0",
     "react-fast-compare": "^2.0.4",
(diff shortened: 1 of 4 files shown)


elliottsj:
Ah, seems we need to export the `Waiter` component along with the `useWait` hook because they must be used together. But TypeScript doesn't allow us to set an arbitrary property on a function.

I've pushed a change which exports both values separately.

PR #488, opened 2019-07-30
useRafLoop

xobotyi:
React hook that calls given function inside the RAF loop without re-rendering parent component if not needed. Loop stops automatically on component unmount.  
Provides controls to stop and start loop manually.


src/useRafLoop.ts
```diff
@@ -0,0 +1,39 @@
+import { DependencyList, useEffect, useRef, useState } from 'react';
+
+export type RafLoopReturns = [() => void, boolean | null, () => void];
```
streamich: I think we can remove `| null` here.

src/useRafLoop.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export default function useRafLoop(callback: CallableFunction, deps?: DependencyList): RafLoopReturns {
+  const raf = useRef<number | null>(null);
+  const [isActive, setIsActive] = useState<boolean | null>(true);
```
streamich: And here we should be able to remove `| null`, too.

src/__tests__/useRafLoop.test.tsx
```diff
@@ -0,0 +1,113 @@
...
+  it('should call a callback constantly inside the raf loop', done => {
+    let calls = 0;
+    const spy = () => calls++;
+    renderHook(() => useRafLoop(spy), { initialProps: false });
+
+    expect(calls).toEqual(0);
+
+    setTimeout(() => {
+      expect(calls).toBeGreaterThanOrEqual(6);
+      expect(calls).toBeLessThan(10);
```
streamich: I'm curious if these tests will always work.

src/__tests__/useRafLoop.test.tsx
```diff
@@ -0,0 +1,113 @@
...
+      done();
+    }, 100);
+  });
+
+  it('second element should represent loop state', done => {
+    let calls = 0;
+    const spy = () => calls++;
+    const hook = renderHook(() => useRafLoop(spy), { initialProps: false });
+
+    expect(hook.result.current[1]).toBeTruthy();
```
streamich: Maybe instead of `.toBeTruthy` we should use `toBe(true)` as "truthy" can be a lot of things, but we will be returning a boolean.

src/useRafLoop.ts
xobotyi: Woopsey =)
Left from first implementation. I'll fix it.

src/__tests__/useRafLoop.test.tsx
xobotyi: Agree.
will fix;

src/__tests__/useRafLoop.test.tsx
xobotyi: More or less.
I've decreasse the lower border to 5, o, for most cases - should pass.

xobotyi:
@streamich i'm developing the react component that provides fully adjustable and stylable scrollbars and have to track the scroll states.
RAF loop is used to run the checker in background when it can be done and not to overload processor with usual loop.

streamich:
@xobotyi I've sent you a [collab invitation](https://github.com/streamich/react-use/invitations).

xobotyi:
@streamich much appreciate

PR #500, opened 2019-08-02
Add tests for state hooks block

Belco90:
Add tests for state hooks block https://github.com/streamich/react-use/issues/101

This include tests for:

- createMemo
- createReducer
- useBoolean
- useCounter
- useNumber
- useGetSet
- useList
- useMap
- useObservable
- usePrevious
- useSetState
- useToggle
- useDefault

Some of them already had tests so I moved them to new tests structure and improve them by adding new test cases.

I also updated react-hooks-testing-library package to scooped version.


src/__tests__/useObservable.test.ts
```diff
@@ -0,0 +1,108 @@
+import { act, renderHook } from '@testing-library/react-hooks';
```
Belco90: Tests for useObservable were showing some console errors. They are fixed although there are some console errors from other tests that I'll fix on corresponding block of tests.

src/__tests__/useMap.test.ts
```diff
@@ -0,0 +1,115 @@
...
+});
+
+it('should get corresponding value for existing provided key', () => {
+  const { result } = setUp({ foo: 'bar', a: 1 });
+  const [, utils] = result.current;
+
+  let value;
+  act(() => {
+    // @ts-ignore
+    value = utils.get('a');
```
Belco90: I'm getting a TS error that I don't even understand so I just flagged it to be ignored.

streamich: If you replace

```ts
const { result } = setUp({ foo: 'bar', a: 1 });
```

by

```ts
const { result } = renderHook(() => useMap({ foo: 'bar', a: 1 }));
```

the error will disappear.

But if you will add `<object>` like

```ts
const { result } = renderHook(() => useMap<object>({ foo: 'bar', a: 1 }));
```

the error will appear again.

It happens because `useMap` hook things that the object is of the type `object` but should be of the type `{foo: string, a: number}`.

---

Alternatively, to fix it you could do

```ts
const setUp = <T extends object>(initialMap?: T) => renderHook(() => useMap(initialMap));

// ...

const { result } = setUp<{ foo: string, a: number }>({ foo: 'bar', a: 1 });
```

Belco90: Cool! I'll fix the ts-ignore flags on my next PR. Improving my TS skills was another motivation to contribute to this repo :)

src/__tests__/useDefault.test.ts
```diff
@@ -0,0 +1,61 @@
...
+import useDefault from '../useDefault';
+
+const setUp = (defaultValue: any, initialValue: any) => renderHook(() => useDefault(defaultValue, initialValue));
+
+describe.each`
+  valueType    | defaultValue | initialValue            | anotherValue
+  ${'number'}  | ${0}         | ${5}                    | ${77}
+  ${'object'}  | ${{}}        | ${{ name: 'John Doe' }} | ${{ name: 'Solid Snake' }}
+  ${'boolean'} | ${false}     | ${false}                | ${true}
+  ${'string'}  | ${''}        | ${'foo'}                | ${'bar'}
```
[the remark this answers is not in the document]
Belco90: Jest is amazing, it has tons of testing utils from different tools! Feel free to ask if you want to know more about something in particular.

PR #504, opened 2019-08-05
feat: useUpsert

PetterIve:
This PR adds the `useUpsert` hook.

`useUpsert` provides a superset of `useList`, with the additional method `upsert`, which combines update or insert functionality. `useUpsert` requires that a comparison function is passed when initialising the hook.

Looking forward to your feedback!

### DIFF
--- a/docs/useUpsert.md
+++ b/docs/useUpsert.md
@@ -1,0 +1,28 @@
+# `useUpsert`
+
+Superset of `useList`. Provides an additional method to upsert (update or insert) an element into the list. 
+## Usage
+
+```jsx
+import {useUpsert} from 'react-use';
+
+const Demo = () => {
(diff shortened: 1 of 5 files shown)


streamich:
What is your use case for this hook?

PetterIve:
I typically find it useful have a list of "entities" (i.e. any identifiable object) which are added to, or updated, through a callback where the entity itself is passed without its list index (e.g. through a standalone form component). The hook basically wraps logic I found myself duplicating between multiple `useList` hooks.

I could make a more thorough example if you'd like.

PS: I'm not really sure if prop drilling an object's index is perforable to this approach, or if there's another way to achieve the same effect..

streamich:
Ah, just noticed it is not in the README, maybe you want to add it on the same line with `useList`.

Issue #513, opened 2019-08-12
[bug] useUpdateEffect is wrong ?


brickspert:
This hook is not working as expected. In the following example, the alert should never popped up since the dependency is never changed. However, for the second time clicking on the Button, the alert popped up.

```
import useUpdateEffect from './index';
export default () => {

  const [state, setState] = useState(0);

  useUpdateEffect(() => {
    alert('execute');
  }, []);

  return (
    <div onClick={() => { setState(state + 1) }}>just for render</div>
  )
}
```

PR #514, opened 2019-08-12
fix: useUpdateEffect run on the wrong time

brickspert:
fix: #513

when the useEffect function change , useEffect will run again，same to deps change~


xobotyi:
LGTM

PR #518, opened 2019-08-13
Dedupe yarn.lock

xobotyi:
Made a `npx yarn-deduplicate` so now `yarn.lock` has no mistakes.
Before this commit there was a mismatch in @types/react versions which lead to `tsc` fail.

Furthermore it became much faster to install.

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,7 +74,7 @@
     "@storybook/react": "5.1.11",
     "@testing-library/react-hooks": "2.0.1",
     "@types/jest": "24.0.17",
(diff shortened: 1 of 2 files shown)


xobotyi:
@streamich or, maybe i shoud only run dedupe for `@types/react`?

xobotyi:
@streamich now `@types/react` version not equals the react's version.
If to simply change the typings version it'll brake tsc build due to multiple react version, so it is requires `yarn.lock` deduplication.

Issue #531, opened 2019-08-19
[bug] rollup build error: Cannot call a namespace ('writeText')


brickspert:
**What is the current behavior?**
build error when i use rollup.

our project is: https://github.com/umijs/hooks/pull/36/files

the error solution like this: https://github.com/rollup/rollup/issues/670

![image](https://user-images.githubusercontent.com/12526493/63236366-048d3700-c270-11e9-8803-f5859597f80a.png)

**What is the expected behavior?**

build success

**A little about versions:**  
- _OS_: mac
- _Browser (vendor and version)_:  chrome
- _React_:  16.8.6
- _Typescript: 3.4.2
- _`react-use`_: 10.6.2
- _Did this worked in the previous package version?_  no

PR #532, opened 2019-08-19
fix: rollup build error: Cannot call a namespace ('writeText')

brickspert:
fix: https://github.com/streamich/react-use/issues/531


brickspert:
the error solution from here: rollup/rollup#670

PR #534, opened 2019-08-19
fix: return from useUpdateEffect

brickspert:
fix useUpdateEffect return missing


brickspert:
This is my fault ~~  so sorry.

Please post as soon as possible~

Issue #308, opened 2019-05-18
useBattery doesn’t work – obsolete API


filipdanic:
The [useBattey hook](https://github.com/streamich/react-use/blob/master/src/useBattery.ts) relies on `navigator.getBattery`, which is an [obsolete API.](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/getBattery) It will work on Chrome, but not many other places.

There doesn’t seem to be anything that works across all browsers at the moment. Which is fine, but I wouldn’t expect a hook to break the entire app because of an implementation detail. 

E.g: https://codesandbox.io/s/sleepy-browser-jf2dg

Comparison:
![use_battery](https://user-images.githubusercontent.com/7009755/57966360-01249f80-7951-11e9-9c1e-46a069066842.png)

At a minimum, it should be noted in the docs. Ideally, the useBattery hook should check if `navigator.getBattery` is working first. It should return something like an object `{supported: false` if `navigator.getBattery` is unsupported.

mqliutie:
Chrome on iOS doesn't support

PR #535, opened 2019-08-19
fix: Reworked useBattery hook

xobotyi:
useBattery now doesn't crashing if used on browsers not supporting `BatteryManager` API';
Fix: #308;

### DIFF
--- a/src/useBattery.ts
+++ b/src/useBattery.ts
@@ -2,6 +2,8 @@ import * as React from 'react';
 import * as isEqual from 'react-fast-compare';
 import { off, on } from './util';
 
+const { useState, useEffect } = React;
+
 export interface BatteryState {
   charging: boolean;
   chargingTime: number;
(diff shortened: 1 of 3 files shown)


src/useBattery.ts
```diff
@@ -1,56 +1,106 @@
...
+    React.useEffect(() => {
+      battery.current && bindBatteryEvents(battery.current);
+
+      return () => {
+        battery.current && unbindBatteryEvents(battery.current);
+      };
+    }, [onChange]);
+
+    React.useEffect(() => {
+      nav.getBattery!().then((bat: BatteryManager) => {
```
streamich: This could probably use `useAsync`, then you also would not need to track `isMounted` state.

src/useBattery.ts
```diff
@@ -1,56 +1,106 @@
...
+export default function useBattery() {
+  const [state, setState] = React.useState<UseBatteryState>({
+    isSupported: nav && typeof nav.getBattery !== 'undefined',
+    level: 1,
+    charging: true,
+    dischargingTime: Infinity,
+    chargingTime: 0,
+  });
+  const battery = React.useRef<BatteryManager>();
+  let isMounted = true;
```
streamich: This should probably be 

```js
const isMounted = useMountedState();
```

src/useBattery.ts
```diff
@@ -1,56 +1,106 @@
...
+
+  if (state.isSupported) {
+    const onChange = React.useCallback(() => {
```
streamich: If you would inline `onChange` function inside `useEffect` you would not need to use `useCallback`.

```js
const data = useAsync();
useEffect(() => {
  if (!data.value) return;
  const onChange = () => {};
}, [data.value]);
```

src/useBattery.ts
```diff
@@ -1,56 +1,106 @@
...
+    isSupported: nav && typeof nav.getBattery !== 'undefined',
+    level: 1,
+    charging: true,
+    dischargingTime: Infinity,
+    chargingTime: 0,
+  });
+  const battery = React.useRef<BatteryManager>();
+  let isMounted = true;
+
+  if (state.isSupported) {
```
streamich: Instead of having this `if` statement, can we instead write this hook assuming that it is client side and Battery API is supported, and then at module scope make a check.

```
const useBattery = isClient && isSupported ? useBatteryClient : useBatteryMock;
export default useBattery;
```

xobotyi: Well, this one is really good idea =)

src/useBattery.ts
xobotyi: Here is more optimal to use a single `useEffect`.

src/useBattery.ts
xobotyi: I don't want the component to re-render on async state change - only on battery state change.
Furthermore i still have to clear bound event handler on unmount so still have to use useEffect where i can combine all things.
So for this case, in terms of performance optimisation - better not to use useAsync hook (IMO)

streamich:
@xobotyi please tell me what you think. I inlined `handleChange` inside `useEffect`, so as to not use `useCallback`.

xobotyi:
@streamich yaeh, now i see it could been better =)
the only thing - i'd left the navigator as separate variable as it was (it is SSR compatible.)

streamich:
I don't understand how will work on the server. I get `ReferenceError`, if I run this:

image

xobotyi:
Yikes! My bad, i'll fix it.

PR #536, opened 2019-08-19
Improve useWindowSize performance with requestAnimationFrame

wardoost:
Use [requestAnimationFrame](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame) to improve performance of resize handler. Currently tests are failing though so it would be nice if someone could point me in the right direction on how to mock requestAnimationFrame 🙂

### DIFF
--- a/src/__tests__/useWindowSize.test.tsx
+++ b/src/__tests__/useWindowSize.test.tsx
@@ -1,5 +1,23 @@
 import { act, renderHook } from '@testing-library/react-hooks';
+import { replaceRaf } from 'raf-stub';
 import useWindowSize from '../useWindowSize';
+
+interface RequestAnimationFrame {
+  reset(): void;
+  step(): void;
+}
+
+declare var requestAnimationFrame: RequestAnimationFrame;
+
+replaceRaf();
+
+beforeEach(() => {
+  requestAnimationFrame.reset();
+});
+
+afterEach(() => {
+  requestAnimationFrame.reset();
+});
 
 // simulate window resize
 function fireResize(type, value) {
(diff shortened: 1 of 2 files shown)


xobotyi:
Dont get the sense of RAF here.
There are no heavy calculations inside so no need in RAF. It shoud depend on developer's implementation IMO.

streamich:
@wardoost 

Maybe you could use `window.requestAnimationFrame` instead of `requestAnimationFrame` and then simply

```js
window.requestAnimationFrame = jest.fn();
```

You could also export it from some module and mock the module instead

```js
export const raf = isClient ? requestAnimationFrame : null;
```

Maybe @Belco90 has ideas.

---

@xobotyi 

I guess the idea is to not trigger component re-render more than once per `requestAnimationFrame` if somebody quickly resizes the screen.

Belco90:
The requestAnimationFrame is hard to test! I'm still pending to create new PR for animation hooks block tested, but I already pushed `useRaf` hook tests, so here you can find how I did it: https://github.com/Belco90/react-use/commit/5bc14d1a302773f1f987dec438e46472c54219ce

There are 2 different things I'm doing on my tests, you only need the first one but I'll explain everything just to clarify

- I installed `raf-stub` to mock the whole requestAnimationFrame and be able to fully control next step or even reset when I need it, so feel free to reinstall it in your branch again and replicate the config and small mocked raf behavior I implemented (and then I'll merge it properly on mine). You can see I'm resetting the raf mock on both beforeEach and afterEach so you get raf in proper status when starting each test, this is very important (tho only the latter should be necessary, I want to make sure it doesn't affect other tests).
- I'm also using `jest.useFakeTimers()` before each test and `jest.clearAllTimers()` after each test because you can pass a delay to the hook so it waits for delay to be reached before start everything. That made my tests more complex, but you don't need that. Fake timers is a jest util that allows you to fully control timers as intervals or timeouts, so with that I can also fully control the status of the timeout to check raf stuff is not called if delay is not reached etc

I hope this helps and don't hesitate to ask me anything about how to implement your tests here!

wardoost:
@Belco90 Thanks for the info, will update this PR soonish 🙂

wardoost:
Sorry for the delay on this one, sometimes real life just gets a bit in the way of code life 🙂

@Belco90 I mainly followed your `useRaf` tests

PR #537, opened 2019-08-19
feat: useDebounce add cancel

brickspert:
In some scenarios，linke [Select search](https://ant.design/components/select/#components-select-demo-select-users)，we want to cancel debounce when onBlur，because it's useless。

so i think cancel is necessary.

our code: https://github.com/umijs/hooks/blob/master/src/useSearch/index.ts#L56


xobotyi:
Removed the `fn.bind(null, args)` because it is basically useless (all theese variables available from the context function been created)

xobotyi:
Tomorrow will write some tests

xobotyi:
@streamich done tests, now it's ready to merge, all good now.
but first have to merge #568

PR #544, opened 2019-08-20
fix: update createReducer to fix initial state

DmacMcgreg:
Incrementing will give NaN

CodeSandbox: https://codesandbox.io/s/condescending-bohr-9xe85


atymchuk:
I came across the same issue.  Thumbs up to merge.

PR #549, opened 2019-08-21
feat: useTimeout add cancel and reset handles

xobotyi:
useTimeout:
- added cancel method;
- added reset method;
- improved docs;
- added tests;

### DIFF
--- a/src/useTimeout.ts
+++ b/src/useTimeout.ts
@@ -1,19 +1,10 @@
-import { useEffect, useState } from 'react';
+import useTimeoutFn from './useTimeoutFn';
+import useUpdate from './useUpdate';
 
-const useTimeout = (ms: number = 0) => {
-  const [ready, setReady] = useState(false);
+export type UseTimeoutReturn = [() => boolean | null, () => void, () => void];
 
-  useEffect(() => {
-    const timer = setTimeout(() => {
-      setReady(true);
-    }, ms);
+export default function useTimeout(ms: number = 0): UseTimeoutReturn {
+  const update = useUpdate();
 
-    return () => {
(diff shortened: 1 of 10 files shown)


xobotyi:
Do not merge - tomorrow i'll change implementation onto `useTimeoutFn` usage.

src/useTimeout.ts
```diff
@@ -1,19 +1,10 @@
...
-    return () => {
-      clearTimeout(timer);
-    };
-  }, [ms]);
-
-  return ready;
-};
-
-export default useTimeout;
+  return useTimeoutFn(update, ms);
```
streamich: Nice!

src/useTimeoutFn.ts
```diff
@@ -0,0 +1,29 @@
...
+  }, [ms, fn]);
+  const clear = useCallback(() => {
+    ready.current = null;
+    timeout.current && clearTimeout(timeout.current);
+  }, []);
+
+  useEffect(() => {
+    set();
+
+    return clear;
```
streamich: Is it possible to inline `set` and `clear` here, so that we don't need to use `useCallback` hook?

src/useTimeout.ts
xobotyi: `useTimeoutFn` is in another PR: https://github.com/streamich/react-use/pull/540

src/useTimeoutFn.ts
xobotyi: Dont think so, they have to:
a) be returned;
b) set should change on function change and delay change;

Thus i dont see possibilities to inline them in useEffect.

PR #551, opened 2019-08-22
feat: add usePreviousDistinct

### LINKED ISSUE #476: usePrevious only tracks the previous of the last render and is affected by other hooks
If i have other hooks in a component that are unrelated to what the `usePrevious` is tracking, some update that causes a rerender from that hook will sync the previous value to the current value. To see an example of this, here's a [sandbox](https://codesandbox.io/s/react-use-useprevious-0s9po). To see the behavior, click `+` a few times and then click `toggle on` to toggle unrelated state. Not that the value for `prev` will update.

In this sandbox I've implemented an alternative usePrevious that tracks a more accurate version of previous.

I can make a PR but wanted to know if maybe we call this new version something else to support the previous version of `usePrevious`.

I understand that `usePrevious` tracks the previous render value, but this version may be more strictly useful in an application. Maybe `usePreviousDistinct`?

psachs21:
Fixes #476

### DIFF
--- a/src/__tests__/usePreviousDistinct.test.tsx
+++ b/src/__tests__/usePreviousDistinct.test.tsx
@@ -1,0 +1,62 @@
+import { renderHook } from '@testing-library/react-hooks';
+import usePrevious from '../usePreviousDistinct';
+
+describe('usePrevious with default compare', () => {
+  const hook = renderHook(props => usePrevious(props), { initialProps: 0 });
+
+  it('should return undefined on initial render', () => {
+    expect(hook.result.current).toBe(undefined);
+  });
+
+  it('should return previous state only after a different value is rendered', () => {
+    expect(hook.result.current).toBeUndefined();
+    hook.rerender(1);
+    expect(hook.result.current).toBe(0);
+    hook.rerender(2);
+    hook.rerender(2);
+    expect(hook.result.current).toBe(1);
+
+    hook.rerender(3);
+    expect(hook.result.current).toBe(2);
+  });
+});
+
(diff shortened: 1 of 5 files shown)


src/__tests__/usePreviousDistinct.test.tsx
```diff
@@ -0,0 +1,64 @@
+import { cleanup, renderHook } from 'react-hooks-testing-library';
+import usePrevious from '../usePreviousDistinct';
+
+afterEach(cleanup);
```
Belco90: Cleanup not necessary anymore, so please remove this line

src/__tests__/usePreviousDistinct.test.tsx
```diff
@@ -0,0 +1,64 @@
+import { cleanup, renderHook } from 'react-hooks-testing-library';
```
Belco90: react-hooks-testing-library was updated to scoped version, so you need to import it from @testing-library/react-hooks

psachs21: Ah, haven't updated to latest in a bit. Will do so.

Belco90:
Looks good to me! Let's see what @xobotyi @streamich think.

xobotyi:
LGTM too

psachs21:
> Nice hook! It looks good to me, but it would need an example file for storybook and some changes on test file.
> 
> EDIT: sorry, just realize the example is already there 💯

No, this was a good catch. I realize i hadn't even run storybook to check if the docs were picked up. I've push an update with the docs.

PR #554, opened 2019-08-22
Animation block tests

Belco90:
Add tests for animation hooks block #101 
This one was hard to test, specially useRaf and useSpring but I learnt a lot testing them!

This include tests for:

- useRaf
- useSpring
- useTimeout
- useUpdate
- useTimeoutFn (already done by someone else! I just adapt it to have same structure as others)

The PR also includes some bug fixes!


src/__stories__/useSpring.story.tsx
```diff
@@ -1,6 +1,6 @@
 import { storiesOf } from '@storybook/react';
 import * as React from 'react';
-import { useSpring } from '..';
+import useSpring from '../useSpring';
```
Belco90: Fix for useSpring demo, it wasn't working.

src/__tests__/useGetSetState.test.ts
```diff
@@ -100,8 +100,7 @@ it('should log an error if set with a patch different than an object', () => {
   const [, set] = result.current;
   expect(mockConsoleError).not.toHaveBeenCalled();
 
-  // @ts-ignore
```
Belco90: Removing ts-ignore flag from my previous PR! You'll find few of these.

src/__tests__/useRaf.test.ts
```diff
@@ -0,0 +1,153 @@
+import { act, renderHook } from '@testing-library/react-hooks';
+import { replaceRaf } from 'raf-stub';
+import useRaf from '../useRaf';
+
+/**
+ * New requestAnimationFrame after being replaced with raf-stub for testing purposes.
+ */
+interface RequestAnimationFrame {
```
Belco90: On this test I'm completely mocking RAF and simulating only the minimum behavior I need.

src/__tests__/useSpring.test.ts
```diff
@@ -0,0 +1,119 @@
+import { act, renderHook } from '@testing-library/react-hooks';
+import useSpring from '../useSpring';
+import { Spring } from 'rebound';
+
+// simulate Spring for testing
+const mockSetCurrentValue = jest.fn();
```
Belco90: Mocking and reimplementing the whole rebound.Spring. This implies that I'm testing implementation here rather than behavior, but it's little bit crazy to test the whole behavior as 1) it's quite difficult and 2) I would be re-testing Spring itself, that's why I'm focusing on mocking and reimplementing everything here.

src/__tests__/useTimeout.test.ts
```diff
@@ -2,137 +2,135 @@ import { act, renderHook, RenderHookResult } from '@testing-library/react-hooks'
 import { useTimeout } from '../index';
 import { UseTimeoutReturn } from '../useTimeout';
 
-describe('useTimeout', () => {
```
Belco90: Just removing the outer describe block as 1) it doesn't add any functionality and 2) it adds unnecessary nested level, but this is actually a personal preference so let me know if you prefer me to stop doing this, or conversely you are happy with this decision (and we could even add some guidelines to the repo about what format should the test follow).

src/__tests__/useTween.test.ts
```diff
@@ -0,0 +1,56 @@
+import { renderHook } from '@testing-library/react-hooks';
```
Belco90: Similar to useSpring tests: I'm mocking useRaf hook and easing fn and checking everything is init properly, as it's actually what this hook is doing (though again this is testing implementation rather than behavior but I think it's fine for this one).

src/useSpring.ts
```diff
@@ -1,31 +1,37 @@
...
 
 const useSpring = (targetValue: number = 0, tension: number = 50, friction: number = 3) => {
   const [spring, setSpring] = useState<Spring | null>(null);
   const [value, setValue] = useState<number>(targetValue);
 
-  useEffect(() => {
-    const listener = {
+  // memoize listener to being able to unsubscribe later properly, otherwise
+  // listener fn will be different on each re-render and wouldn't unsubscribe properly.
+  const listener = useMemo(
```
Belco90: I found couple of issues on this one by adding tests! First one: listener has to be defined out of useEffect and memoized, not for performance but for keeping same function within listener const so it can be referenced properly when unsubscribing (otherwise, as explained on the code comment, listener fn never would be the same as it generates a different one on each update).

src/useSpring.ts
```diff
@@ -1,31 +1,37 @@
...
     return () => {
-      spring.removeListener(listener);
-      setSpring(null);
+      if (spring) {
+        spring.removeListener(listener);
+        setSpring(null);
+      }
     };
-  }, [tension, friction]);
+  }, [tension, friction, spring]);
```
Belco90: Second issue: spring wasn't added to deps dependency so useEffect wasn't updated its callback to reference proper spring. What does that imply? Listener was never removed.

I usually work with vanilla js and I always set up this [ESLint plugin](https://www.npmjs.com/package/eslint-plugin-react-hooks) to get proper validation on hooks deps. This plugin would have detected this (or at least warned you) but sadly it's not available yet for TSLint: I found [this equivalent plugin](https://www.npmjs.com/package/tslint-react-hooks) but it doesn't implement the deps check yet (tho [it has an open issue for it](https://github.com/Gelio/tslint-react-hooks/issues/9), so I hope we can get it soon for TSLint)

streamich: I think `react-use` has to switch to ESLint in foreseeable future, as TSLint is deprecated.

streamich:
@Belco90 I've sent you [a collab invite](https://github.com/streamich/react-use/invitations).

Belco90:
Awesome! Thanks for the collab invitation.

PR #556, opened 2019-08-23
Removed prepended window from setTimeout

mitchheddles:
Similar to the `clearTimeout`, the `setTimeout` shouldn't need the `window` object.
As a result this function can work in node environment, e.g. SSR.


xobotyi:
But this it should fail tsc check:expressionless:

mitchheddles:
> But this it should fail tsc check😑

Fixing now, my bad.

xobotyi:
LGTM now.

Issue #555, opened 2019-08-23
useSetState infinite loops when following React Hooks Best Practices Guide


TylerR909:
According to the Best Practices described by [React-Hooks FAQ](https://reactjs.org/docs/hooks-faq.html#is-it-safe-to-omit-functions-from-the-list-of-dependencies) and its associated ESLint [react-hooks/exhaustive-deps](https://www.npmjs.com/package/eslint-plugin-react-hooks) plugin, when `setState` is used inside an effect, it should be included in the list of dependencies.
 
**What is the current behavior?**

This causes an infinite loop where `setState` is called from within an effect and a different `setState` is returned from the next `useSetState` hook.

**Steps to reproduce **

https://codesandbox.io/embed/practical-flower-4jpwp

**What is the expected behavior?**

Some sort of smart caching, useMemo, or useCallback around the `setState` generator that depends on the `useState`'s `setState`. Basically a person diligently following the React-Hooks Best Practices should be able to use setState comfortably inside a useEffect without causing infinite rerenders.

**A little about versions:**  
- _OS_: OSX
- _Browser (vendor and version)_: Chrome 76.0.3809.100
- _React_: 16.8.6
- _`react-use`_: 9.0.0 and 11.0.0
- _Did this worked in the previous package version?_ No

xobotyi:
fixed with v11.0.2

PR #557, opened 2019-08-23
fix(useSetState): memoize setState callback

TylerR909:
Addresses https://github.com/streamich/react-use/issues/555 

Wraps the callback from `useSetState` in `useCallback` to memoize it so that folks following the Best Practices from the [React FAQ](https://reactjs.org/docs/hooks-faq.html#is-it-safe-to-omit-functions-from-the-list-of-dependencies) with the `react-hooks/exhaustive-deps` [rule](https://www.npmjs.com/package/eslint-plugin-react-hooks) enabled don't trigger endless rerenders in a `useEffect(() => {}, [setState]);` where `setState` is constantly changing out from under them.

I tried to write a couple tests for this but I'm not really sure what angle to approach it from or how to layer everything together to test for something useful.

### Problematic code fixed by this PR:

```js
const myComponent = () => {
  const [{ loading, data }, setState] = useSetState({ loading: true, data: null });

  useEffect(() => {
    console.log('hi');
    // fetch data
    setState({ loading: false, data: 'data!' });
  }, [setState])

  return null;
}
```

image

### DIFF
--- a/src/__tests__/useSetState.test.ts
+++ b/src/__tests__/useSetState.test.ts
@@ -40,3 +40,21 @@ it('should merge changes into current state when providing function', () => {
 
   expect(result.current[0]).toEqual({ foo: 'bar', count: 2, someBool: true });
 });
+
+/**
+ * Enforces cases where a hook can safely depend on the callback without
+ * causing an endless rerender cycle: useEffect(() => setState({ data }), [setState]);
+ */
+it('should return a memoized setState callback', () => {
+  const { result, rerender } = setUp({ ok: false });
+  const [, setState1] = result.current;
+
+  act(() => {
+    setState1({ ok: true });
+  });
+  rerender();
+
+  const [, setState2] = result.current;
+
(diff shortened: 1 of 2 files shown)


Belco90:
Ey, thanks for your fix! Testing this should be easy: at the end you need to check if the returned `setState` fn is always the same, so you can add an additional test to `useSetState.test.ts` like 'should return always same setter instance' where:

1. render the hook and save its returned `setState` in a const
2. rerender the hook and save new returned `setState` in a new const
3. compare two saved `setState` are the same with jest matcher `toBe` as `expect(Object.is(received, expected)).toBe(true)`

It should fail with the previous code and pass with your new fix. You can find an example about how rerender a hook within `usePrevious.test.ts` file.

TylerR909:
👍 

Added a test to do just that. Validated that it broke with the old code and passes with the new code.

Hope it meets with your approval. Thanks!

PR #559, opened 2019-08-23
feat: add useStateList

d-asensio:


xobotyi:
Hook name is extremely unintuitive😵

d-asensio:
@xobotyi I agree 👍 any proposal?

`useStateRotate`
`useListRotate`

😅

xobotyi:
`useArrayIterate`
`useListIterate`

d-asensio:
mmm... But the term 'iterate' is missing the important part of this hook... it returns to the first element when reaches the last element of the array. So it is a [circular shift](https://en.wikipedia.org/wiki/Circular_shift).

`useCircularShift`??

d-asensio:
Maybe `useCircularIterate` is more appropriate since we are not shifting (that implies modification) the array. Just iterating it in a circular way.

xobotyi:
@d-asensio this one looks good.

d-asensio:
@xobotyi done!

streamich:
This is basically a *finite state machine* with *state transitions* being adjacent list elements. Maybe we can call it something like `useStateList`? And change interface to:

```js
const { state, next, prev } = useStateList({ list, circular: true });
```

d-asensio:
@streamich Yes! That's the idea! I use it as a _finite state machine_, specifically to control the animation phases of a [`<Collapsible />` component](https://codesandbox.io/s/collapsiblecomponent-w8xjn).

So I agree with the proposed interface!

I'll submit a PR with the corresponding changes during tomorrow ;)

d-asensio:
@streamich @xobotyi done!

PR #567, opened 2019-08-25
fix: useThrottle & useThrottleFn proper timeout type;

xobotyi:


src/useThrottle.ts
```diff
@@ -3,7 +3,7 @@ import useUnmount from './useUnmount';
 
 const useThrottle = <T>(value: T, ms: number = 200) => {
   const [state, setState] = useState<T>(value);
-  const timeout = useRef<any>(null);
+  const timeout = useRef<ReturnType<typeof setTimeout>>();
```
streamich: This is nice, I'm going to use this in the future.

PR #568, opened 2019-08-26
fix: make useTimeoutFn work properly:

xobotyi:
- reset function should clear previous timeout;
- function change should not reset the timeout;

### DIFF
--- a/src/__tests__/useTimeoutFn.test.ts
+++ b/src/__tests__/useTimeoutFn.test.ts
@@ -123,7 +123,7 @@ describe('useTimeoutFn', () => {
     expect(spy).not.toHaveBeenCalled();
 
     const spy2 = jest.fn();
-    hook.rerender({ delay: 5, cb: spy2 });
+    hook.rerender({ delay: 50, cb: spy2 });
 
     jest.advanceTimersByTime(25);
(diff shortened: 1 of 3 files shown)


src/__tests__/useTimeoutFn.test.ts
```diff
@@ -108,9 +110,23 @@ describe('useTimeoutFn', () => {
...
+    const [spy, hook] = getHook(50);
+
+    jest.advanceTimersByTime(25);
+    expect(spy).not.toHaveBeenCalled();
+
+    const spy2 = jest.fn();
+    hook.rerender({ delay: 5, cb: spy2 });
+
+    jest.advanceTimersByTime(25);
+    expect(spy).not.toHaveBeenCalled();
```
Belco90: One thing I don't fully understand: if it should not reset timout, then shouldn't this mock be called? As timers where advanced 25 + 25 so it reached its timeout, but it wasn't called.

xobotyi: Actually there is a mistake but not in part you pointed =) re-rendered delay was wrong.
Anyways, about your question.

No, if it would reset timeout it should have been called at 75ms, not 50, but it's called on 50, so everything is okay.

PR #572, opened 2019-08-27
Issue 401/use debounce and cancellation

### LINKED ISSUE #401: useDebounce and cancellation
Similar to how [debounce](https://www.npmjs.com/package/debounce) does it, the hook should return callbacks to clear and flush. This is important for scenarios, where some other action has immediate priority and debounce, should be silenced in that case. The typical is having debounced submit of form input but be able to just hit Enter to speed up the action.

dichuvichkin:
Closes #401


xobotyi:
#531 already adds cancelation

Issue #284, opened 2019-05-01
useAudio: paused instead of isPlaying


MajorBreakfast:
Consider adding a `paused` property to the state and to remove the `isPlaying` property

Advantages:
- Symmetry with the HTML element's API
- `isPlaying` isn't precise: It can either mean that the audio is currently playing (which it does not in this case) or I could mean that the audio is currently playing or will play as soon as it loads (which it does in this case). This is probably also the reason why the property has been standardized as `paused`.

Serchinastico:
I'm pretty sure this issue can be closed now 😄

streamich:
yup, thanks

https://github.com/streamich/react-use/blob/master/src/util/createHTMLMediaHook.ts#L14

PR #579, opened 2019-08-30
Using paused property instead of isPlaying

realdennis:
According to this issue https://github.com/streamich/react-use/issues/284 , change from `isPlaying` to precise property name `paused`.

### DIFF
--- a/docs/useAudio.md
+++ b/docs/useAudio.md
@@ -2,15 +2,16 @@
 
 Creates `<audio>` element, tracks its state and exposes playback controls.
 
+
 ## Usage
 
 ```jsx
-import { useAudio } from 'react-use';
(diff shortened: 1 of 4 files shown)


realdennis:
@Belco90 
Thanks for the reminder, I update the corresponding documents, and add some test case for the useAudio custom hook.

But, I still have no idea how to test the `<HTMLMediaElement>` `seek` & `play` in jest dom. (The [thread in stackoverflow](https://stackoverflow.com/questions/51829319/how-to-mock-video-pause-function-using-jest) which discuss testing media element in jest ).

Belco90:
Looks good to me with that, testing html media elements with jest dom is not that easy so just testing the hook we can assume the html media element will work as expected.

Issue #190, opened 2019-03-28
Type error when using useAsync


lvl99:
I'm getting the below error:

<img width="691" alt="image" src="https://user-images.githubusercontent.com/238480/55195452-d74ec800-51ac-11e9-9511-516f549416a2.png">

When looking at the definition for `AsyncState`:

```typescript
export declare type AsyncState<T> = {
    loading: true;
} | {
    loading: false;
    error: Error;
} | {
    loading: false;
    error?: undefined;
    value: T;
};
```

Would this be better expressed as the following?:
```typescript
export declare type AsyncState<T> = {
    loading: boolean;
    error?: Error;
    value?: T;
};
```

I see in the `useAsync` code as well that there's three shapes that `AsyncState` could be. I'm not sure what is best, but I feel that the `AsyncState` should have all properties initialised first and then modified accordingly. At the moment, if I want to test for `AsyncState.value` I'd have to do first `!!AsyncState.loading && AsyncState.value`, and also means that if I was using a hook which is watching the `AsyncState.value` I'd also need to watch `AsyncState.loading`, which again seems cumbersome (and having the `AsyncState.value` available could also be reused for https://github.com/streamich/react-use/issues/172).

lvl99:
Ah, I think I can understand better the architecture now. There's no `value` or `error` due to the async promise which might make the internal `state` object out of date?

lvl99:
Seems that the `useAsync` story has a typescript error as well:
<img width="903" alt="image" src="https://user-images.githubusercontent.com/238480/55198420-36650a80-51b6-11e9-824a-43bd4b4285da.png">

And `useAsyncRetry` story is broken due to `deps` as well, plus I can't actually get it to retry the async call.

I think I might steer clear of this lib until it has unit tests...

xiaoxiangmoe:
Sorry, this is my change that caused your incompatibility error.
Here is my initial thoughts when I made this change:
1. I change AsyncState it because it seems 

```ts
type AsyncState<T> = {
    loading: boolean;
    error?: Error;
    value?: T;
};
```
The interface you modified does not guarantee  and the value of value is not empty when `Error===undefined` and `loading===false`.

2. deps is needed, according to the habit of useEffect.

---

Your suggestion sounds reasonable. 

Maybe I should change it into

```ts
export declare type AsyncState<T> = {
    loading: true;
    error: undefined;
    value: undefined;
} | {
    loading: false;
    error: Error;
    value: undefined;
} | {
[...]

Issue #88, opened 2019-01-02
useApi ?


xialvjun:
just like `Query Mutation in react-apollo`:

```jsx
<Query query={gql_document} variables={variables}>
  {({ data, error, loading }) => <div></div>}
</Query>

<Mutation mutation={gql_document}>
  {(mutation_api, mutation_result) => {
    const { data, error, loading } = mutation_result;
    return <div></div>
  }}
</Mutation>
```

And now we have `useAsync`, but it can only act as `Query`.

We should have an `useApi` to act as `Mutation`.

```jsx
// A very very naive implementation
const useApi = (promisified_api, variables) => {
   const [state, set_state] = useState({loading: !!variables});
   const api = async (...variables) => {
    set_state({loading:true});
    const data = await promisified_api(...variables);
    set_state({data});
  }
  // we should expose `set_state`, it can act as Query's updteQuery, and use can set_state({error:null}) manually when close the error dialog.
  return [api, state, set_state];
}

const [api, state, set_state] = useApi(fn_return_promise, maybe_variables);
```

benneq:
I'd say that there's no need for `useApi`, but instead the `useAsync` hook should get extended. It has some design flaws and is missing some features.

I've written my own `useAsync` version, which works like this:
```
const getData = (id: string) => {
    return getDataByIdViaAPI(id);
}

const TodoTaskEdit: React.FunctionComponent<Props> = (props) => {
    const { isLoading, isResult, isError, result, error, load } = useAsync(getTask);

    useEffect(() => {
        load(props.id);
    }, []);
}
```

It has additional `isResult` and `isError` properties, because it's possible to have a Promise which successfully resolves to `null` or `undefined`.

And it doesn't run your promise automatically. That's why there is the `load` parameter, which is a function, that will run your promise.

Maybe one should then add an `loadOnMount` option, so you don't have to write the `useEffect` hook, which only triggers the `load` function.

Here's my implementation. There may be some optimizations left 😄 
```ts
import { useState } from 'react';

export const useAsync = <T, ARGS extends any[]>(promise: (...args: ARGS) => Promise<T>) => {
    const [state, setState] = useState({
        isLoading: false,
        isResult: false,
        isError: false,
        result: undefined as undefined | T,
        error: undefined
    });

    const load = async (...args: ARGS) => {
        setState({ isLoading: true, isResult: false, isError: false, result: undefined, error: undefined });
        try {
            const result = await promise(...args);
[...]

PR #581, opened 2019-08-30
Improve with AsyncState type definition

Rogdham:
## Issue

The `AsyncState` type could return an object with `loading` being false, but no `value` nor `error`.

This makes handling the `value` returned by e.g. `useAsync` quite difficult:

code

<details><summary>Code used for easy copy/paste</summary>
<p>

```ts
import { useAsync } from 'react-use';

const getPromise = async (): Promise<number> => 6;

const test = (): string => {
  const state = useAsync(getPromise, []);
  if (state.loading === true) {
    return 'loading...';
  }
  if (state.error !== undefined) {
    return state.error.message;
  }

  const result = 7 * state.value;
  return `${result}`;
};

export default test;
```

</p>
</details>

In the above example, the loading and error cases have already been ruled out, so `value` should not be `undefined`, creating issues:
 * This makes developers test if the `value` is undefined or not, which is an useless test;
 * If the `Promise` was actually returning a value that could be undefined, then we have no assurance that the (undefined) value actually comes from the promise.

### Commit creating the issue

In https://github.com/streamich/react-use/pull/348, @NullVoxPopuli merged the type `AsyncState` in `useAsync` with the one in `useAsyncFn`. However, there were different!

```diff
*** useAsync_type.ts
--- useAsyncFn_type.ts
***************
*** 1,7 ****
  export type AsyncState<T> =
    | {
!       loading: true;
        error?: undefined;
        value?: undefined;
      }
    | {
--- 1,7 ----
  export type AsyncState<T> =
    | {
!       loading: boolean;
        error?: undefined;
        value?: undefined;
      }
    | {
```

The fact that `loading` was `true` and not `boolean` solved the issue in `useAsync`, but it is not possible in `useAsyncFn` because of the initial value.

### Possibly related issue

@wardoost 's screenshot in https://github.com/streamich/react-use/issues/190#issuecomment-478305429 seems related even though the issue may not be

## Proposed solution

In this PR, I revert the diff above in the type `AsyncState`, and overload the type of `useAsyncFn` to  make the best of both worlds.

In other words, for the type `AsyncState<T>` if `loading` is `false`, then either `error` (of type `Error`) or `value` (of type `T`) has to be specified.

code2

### Breaking change

Before this PR, it was possible to call `useAsyncFn` with an `initalState` set to `{ loading: false }`, although probably nobody did because it was the default if no `initialState` was provided.

To replicate the behaviour, you can either:
 * use `{ loading: false, value: undefined }` as the `initialState`
 * do not specify `initialState` at all, since it is the default value

I would argue that this breaking change is minor as people in that case would have probably chosen to not specify `initialState` already.


streamich:
In your example, it seems you can replace this line

```ts
const result = 7 * state.value;
```

by this (not `!` non-null assertion):

```ts
const result = 7 * state.value!;
```

would that solve your case?

Rogdham:
> would that solve your case?

Thank you for your reply! That solves the typing issue (an other idea would have been testing if `value` is `undefined`), but does not solve the issue at hand. I would argue that it's more like a workaround against a weak typing of the returned value.

See, as a TypeScript user of `useAsync`, I don't want to look at the way `useAsync` is implemented, I should be able to rely exclusively on docs and types.

 * On one hand, reading [the doc](https://github.com/streamich/react-use/blob/master/docs/useAsync.md), I would have assumed that if `loading` is `false` and `error` is `undefined`, then the `value` is returned directly.
 * On the other hand, looking at the returned types let me think that there is a case in which `useAsync` returns an object `{ loading: false, error: undefined, value: undefined }`, and as a result I should test if `value` is `undefined` as well (but when that case arises and what to do in that case is unclear).

So I would say that the docs and the typing disagree, which is confusing for the users of `useAsync`: which one to believe?

To conclude with, for me this is a bug in the typing, so that's why I'm submitting a PR to fix it :relaxed:

---

_Side note: I have updated my code to have better types for `AsyncFn` as well (+ rebased on master)._

s-thom:
I came across this issue as well, and ended up writing my own type definition that I just cast the result to.

```ts
type AsyncState<T> =
  | {
      loading: true;
    }
  | {
      loading: false;
      error: Error;
      value: undefined;
    }
  | {
      loading: false;
      error: undefined;
      value: T;
    };
```
The main difference between this and your solution is the loading state doesn't have `error` or `undefined`, which better matches the implementation. I've also set them to undefined in the opposite cases, which means the compiler doesn't yell at you when you do `state.value`.

Rogdham:
> The main difference between this and your solution is the loading state doesn't have `error` or `undefined`, which better matches the implementation.

The reason there is, is so that we could do type destructuring:
```ts
const { loading, error, value } = useAsync(…);


```

> I've also set them explicitly to undefined, because `?: undefined` looks weird to me

This is to allow the following: `{ loading: false, value: 42 }` without specifying `error: undefined` for example.

s-thom:
Huh, I hadn't considered those. I generally avoid destructuring (just personal taste, but I have been using t a bit more lately), but the `?:` use case is just something I hadn't considered.

xobotyi:
IMO there should be one more state falg like `loaded` or loading should be renamed to `loaded` and has `boolean | null` type, where:
* `null` - loading not started
* `false` - loading strated but not done
* `true` - loading done

Because current state signature has ambiguous `loaded: false` state meaning both, not started and done states;

Rogdham:
@streamich any thought on @xobotyi 's comment?

If I understand correctly the proposition: returned values of `useAsyncFn`:
 * initial value (assuming no `initialState` argument`):
```ts
{
  loaded: false,
  loading: false,
}
```
 * once the callback has been called:
```ts
{
  loaded: false,
  loading: true,
}
```
 * if the callback resolves:
```ts
{
  loaded: true,
  loading: false,
  value: 42,
}
```

 * if the callback rejects:
```ts
{
  loaded: true,
  loading: false,
  error: new Error('Something went wrong'),
}
```

wardoost:
@xobotyi, @Rogdham Only adding a boolean for `loaded` is not enough, a promise can also throw `undefined`. We could add an `errored` key to fix this, which makes the hook's API bigger than it needs to be. It's what is proposed in https://github.com/streamich/react-use/issues/88#issuecomment-458964715 but with different a different signature.

A status key which indicates all these different states could also work, see https://github.com/streamich/react-use/issues/88#issuecomment-458979326. This would be a breaking change but seems like the most complete solution to me.

Feel free to open a PR for this or continue the conversation in https://github.com/streamich/react-use/issues/88 as this Pr is about type definitions.

src/useAsyncFn.ts
```diff
@@ -18,17 +18,33 @@ export type AsyncState<T> =
...
 
-export type AsyncFn<Result = any, Args extends any[] = any[]> = [
-  AsyncState<Result>,
+export type AsyncFn<Result = any, Args extends any[] = any[], InitialState = Result> = [
+  AsyncState<Result | InitialState>,
   (...args: Args | []) => Promise<Result>
 ];
 
-export default function useAsyncFn<Result = any, Args extends any[] = any[]>(
+// initialState with value: returns AsyncFn<Result, Args>
```
wardoost: Forgot to remove first 2 `useAsyncFn` function definitions?

wardoost: Forgot to remove first 2 useAsyncFn function definitions?

src/useAsyncFn.ts
```diff
@@ -18,17 +18,33 @@ export type AsyncState<T> =
       value: T;
```
wardoost: Couldn't we just do `value: T | undefined;` on this line so we don't need to define an `InitialState` for the `AsyncFn` type and don't need to defined 2 version of `AsyncState` for the `initialState` parameter? A Promise can resolve `undefined` anyways 🤷‍♂️

xobotyi:
@wardoost same as throw undefined it can fulfill with undefined %)
But its still be a 'loaded:true' + 'loading:false', this one is non-breaking change that solves the problem, despite the moment that I'd done as in comment you linked if I develop that hook.

wardoost:
@xobotyi So we would have to add both a `loaded` and `errored` key to be able to determine if a promise throwed or resolved an `undefined` value as pointed out before.

It's easy to say "I would have done if I did it" especially considering the original was written back in October 2018 when hooks wasn't even released in the stable version of React. It's always best to be careful about introducing breaking changes but I think have a single status key should be up for consideration because if keeps the hook's API more clear despite being a breaking change. We could also add a `status` or `state` key to the state with `'idle' | 'pending' | 'rejected' | 'fulfilled'` as possible values and keep the loading key so it's not a breaking change, add a dev warning when the `loading` key is used and introduce the breaking change at a later change to give devs time to update existing code.

I think we best continue the conversation on this in #88 as this PR is about type definitions.

src/useAsyncFn.ts
wardoost: Didn't see this was function overloading, not using TS enough in daily work 😬

src/useAsyncFn.ts
Rogdham: I agree that Promise can resolve to `undefined`, but in that case `undefined` will be _inside_ `T`.

If we allow `value: T | undefined`, then developers will need to check against `undefined`, i.e. an problem similar to what this issue is about:

Rogdham:
> Only adding a boolean for loaded is not enough, a promise can also throw undefined.

I did not considered that for the following reasons:
 1. The existing type already uses `Error` for the `error` key (which will not work if a Promise rejects with a number for example);
 2. I use the [`no-throw-literal` ESLint rule](https://eslint.org/docs/rules/no-throw-literal) which checks that Promises rejects `Error`s: _”It is considered good practice to only throw the Error object itself or an object using the Error object as base objects for user-defined exceptions.”_ I'm not sure if this is a well-established practice, or if it's ok to assume that in any way.

What do you think @wardoost? Especially concerning the point 1 above, i.e. existing type already asserts that Promise reject `Error`s.

streamich:
__TLDR;__ maybe the types could be something like this:

```ts
export type AsyncState<T, E = Error> =
  | {
      state: 'pending';
      loading: true;
      error?: unknown;
      value?: unknown;
    }
  | {
      state: 'rejected';
      loading: false;
      error: E;
      value?: unknown;
    }
  | {
      state: 'fulfilled';
      loading: false;
      error?: unknown;
      value: T;
    };

export type AsyncStateFn<T, E = Error> =
  | {
      state: 'idle';
      loading?: unknown;
      error?: unknown;
      value?: unknown;
    }
  | AsyncState<T, E>;
```

---

1. The initial API for this hook was probably inspired by Apollo Client, [which returns `{ loading, error, data }`](https://www.apollographql.com/docs/react/data/queries/#executing-a-query), this hook swapped `data` for `value`, [which led to `{ loading, error, value }` tuple](https://github.com/streamich/react-use/commit/258d6964388c062b6bc0ee8db39d6d29eb072901#diff-8e6e0808e6745d8058a2f9cbd476172dR4-R6), which I believe still should work fine for the `useAsync` hook (possibly with some tweaks). It is the `useAsyncFn` hook that is more complex.

2. The original `useAsync` hook (and still now) executes the asynchronous function "immediately", which means that `{ loading, error }` should be enough to capture the state of the asynchronous function. Eventually, `useAsyncFn` hook was added, which does not execute the asynchronous function immediately, but instead user can choose when to call it. This adds extra state, which currently, I believe, is not fully captured by the `{ loading, error }` tuple. Hence, for example @xobotyi, has proposed to add an extra `{ loaded }` state.

3. Currently, `useAsync` uses `useAsyncFn` hook internally, I wonder if `useAsyncFn` should use `useAsync` instead, or maybe even the two hooks could be separate.

4. The error property does not necessarily need to be of type `Error`, maybe that's why in the original implementation it was `Error | any`. We should still allow for the possibility of `error` not being of type `Error`, maybe using `Error` as the default type, but allowing user to overwrite it, like `const useAsync = <E = Error>() => { ... }`.

5. Another case which current `useAsync` implementation does not really cover (as mentioned by @wardoost  and others) is that if `error` itself is of type `undefined`. Imagine the asynchronous function throws `undefined`, the state would be `{ loading: false, error: undefined, value: undefined }`. Did promise *resolve* or *reject*? One could, of course, check if the `undefined` values on `error` and `value` are own-properties or missing keys, but that is 🤯. So, even the state of `useAsync` hook today probably does not capture correctly all cases.

6. Probably `{ loading, error }` captures the state of the operation correctly in the Apollo Client, as their `error` is always of type `ApolloError`, but in `react-use` we need to deal with all possible cases. Hence maybe we could [add a more specific `state` property](https://github.com/streamich/react-use/issues/88#issuecomment-458979326), which would *precisely identify* the state; `error` and `value` would not be used for state identification&mdash;those would just hold data. Also, we could still keep `loading` (but deprecated), so it wold not be a breaking change (like suggested by @wardoost and possibly name it `status`, not `state`).

7. As mentioned by @Rogdham, we should keep all properties in all unioned interfaces that create the `AsyncState` type, so we can do destructuring `const { loading, value, error } = useAsync(...)`.

8. Maybe we could add `onFinally` effect [like suggested here](https://github.com/streamich/react-use/issues/88#issuecomment-466614687).

9. There is `useAsyncRetry` hook, but no `useAsyncFnRetry`; I wonder if we need one.

PR #585, opened 2019-09-03
test:add test for useThrottle and useThrottleFn

lintuming:
@streamich Hi! I try add a test for `useThrottleFn`,can you guys take a look for it?

### DIFF
--- a/src/__tests__/useThrottle.test.ts
+++ b/src/__tests__/useThrottle.test.ts
@@ -51,9 +51,7 @@ describe('useThrottle', () => {
       expect(hook.result.current).toBe(1);
       done();
     });
-    jest.advanceTimersByTime(100);
-    expect(hook.result.current).toBe(0);
-    jest.advanceTimersByTime(100);
+    jest.advanceTimersByTime(200);
   });
 
   it('should not update the value after the given time', () => {
--- a/src/__tests__/useThrottle.test.ts
+++ b/src/__tests__/useThrottle.test.ts
@@ -1,0 +1,16 @@
+import { renderHook, act } from '@testing-library/react-hooks';
+import useThrottle from '../useThrottle';
+
+it('should throttle the given value by specific time', () => {
+  jest.useFakeTimers();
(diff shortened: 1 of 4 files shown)


src/__tests__/useThrottle.test.ts
```diff
@@ -0,0 +1,16 @@
...
+  const { result, rerender, unmount } = renderHook(props => useThrottle(props, 1000), { initialProps: 0 });
+  expect(result.current).toBe(0);
+  rerender(1);
+  expect(result.current).toBe(0);
+  act(() => jest.advanceTimersByTime(1000));
+  expect(result.current).toBe(1);
+  act(() => jest.advanceTimersByTime(1000));
+  rerender(2);
+  act(() => jest.advanceTimersByTime(500));
+  unmount();
```
Belco90: Can you split this single test into different ones? So it's more clear what you are testing on it, like "should clear pending timeouts on unmount". Also it would be interesting to test additional cases like rerendering the component with different ms prop or check the result is same as before if time is advanced less than expected ms.

src/__tests__/useThrottleFn.test.ts
```diff
@@ -0,0 +1,32 @@
+import { renderHook, act } from '@testing-library/react-hooks';
```
Belco90: Same comments for this one as the other test file.

src/__tests__/useThrottle.test.ts
```diff
@@ -0,0 +1,16 @@
+import { renderHook, act } from '@testing-library/react-hooks';
+import useThrottle from '../useThrottle';
+
+it('should throttle the given value by specific time', () => {
+  jest.useFakeTimers();
```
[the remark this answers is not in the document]
lintuming: Thanks! I will change that part when i have time.

lintuming:
@Belco90 Hi!I'v changed thoese parts according to your requested changes,please take a look for it when you have time.

PR #586, opened 2019-09-03
Improve useThrottleFn type definition

Rogdham:
## Issue

The typing of `useThrottleFn(fn, ms, args)` does not make sure that the type of `args` matches the type of the arguments of `fn`.

This may lead to runtime issues, for example with the following code:

```tsx
import React, { FC } from 'react';
import { useThrottleFn } from 'react-use';

interface Rect {
  size: {
    width: number;
    height: number;
  };
  color: string;
}

const area = (rect: Rect): number => rect.size.width * rect.size.height;

const Test: FC<{ foo: string }> = ({ foo }): JSX.Element => {
  const bar = useThrottleFn(area, 200, [foo]);
  return <div>{bar}</div>;
};

export default Test;
```

The above will crash at runtime, since `foo` (a `string`) is passed to `area` instead of a `Rect`:

error

## Proposed solution

After this PR, the type is properly checked, and typing errors are raised appropriately:

code

### DIFF
--- a/src/useThrottleFn.ts
+++ b/src/useThrottleFn.ts
@@ -1,7 +1,7 @@
 import { useEffect, useRef, useState } from 'react';
 import useUnmount from './useUnmount';
 
-const useThrottleFn = <T>(fn: (...args: any[]) => T, ms: number = 200, args: any[]) => {
+const useThrottleFn = <T, U extends any[]>(fn: (...args: U) => T, ms: number = 200, args: U) => {
   const [state, setState] = useState<T>(null as any);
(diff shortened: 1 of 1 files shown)


Rogdham:
Thinking back on this, it seems that there is an other issue: the first returned value is `null` due to the `const [state, setState] = useState<T>(null as any);`. I have some runtime crash because of this.

Note that this issue is hidden in the type because `null` is cast as `any`, so the returned value of `useThrottleFn` is inferred of type `T` instead of `T | null`.

Should I open a distinct issue for this?

PR #588, opened 2019-09-04
feat: useStateValidator

xobotyi:
Handy for simple forms field validation;

The only thing i can't decide - should it automatically revalidate state on validator change or not.

Current impelemtation uses auto-revalidation, so it forces developer to wrap validator in `useCallback` by hand to avoid cycluc re-renders.
The other way is to add one more argument to the end something like `revalidateDeps` and wrap validator into `useCallback` automatically using these deps.

@streamich @Belco90 what do you think?


xobotyi:
@Belco90 slept this off and now im sure it's better to leave it outside, because validator can be a single function for many fields and forced `useCallback` wrapping will cause worse memory-wise performance.

Belco90:
I guess this approach even makes more sense than the other, as the validator itself shouldn't change that often or not even change at all. Would make sense to actually implement both hooks?

Issue #456, opened 2019-07-12
useCallback usage


AjaxSolutions:
Why the get, set, remove and remove functions are not wrapped in useCallback?

https://github.com/streamich/react-use/blob/master/src/useMap.ts

The same question applies to functions returned by useList.

streamich:
What is your use case?

Granipouss:
Hi,

I tried to use `useMap`  to hold values of a form and the fact that the methods were not memoized basically made the form unusable due to constant rerenders.

IMO, it would be a good idea to propose the most stable (i.e. memoized) return values possible. Similar to the way `useState` and `useReducer` setter and dispatcher are memoized.

Cheers,

Belco90:
I agree with @Granipouss: ideally all these returned callbacks should be memoized to always return same instance if same args provided. We have fixed few of them, in previous PRs like this one https://github.com/streamich/react-use/pull/570

@AjaxSolutions  could you create a PR for fixing those callbacks on `useCallback` and `useMap` hooks? Thanks.

xobotyi:
Looked inside the hook and scared 🙈
TBH there is almost everything bad about it, starting from non-memoized methods, ending with object recreation each insert-delete.

IMO it should use ref to a storage value, thus it can have the maximum performance and the only callbacks instance with no need to update them ever.
@Granipouss @Belco90 @streamich what do you think? If everyone agree i'd like to take it. Remastering the code would be pretty simple for this one.

Belco90:
Callbacks returned should be memoized, but I'm not so sure the actual object should be memoized too. You need to return a new object when updated, otherwise the component using this hook won't know when the object has been updated. Think about `React.setState` using an object: everytime you set something on it, it returns a new object so you can _react_ to changes in your component. I don't know if I misunderstood what you meant.

xobotyi:
@Belco90 the component re-rendering can be done with a single `useUpdate` hook whose call can be done in handles (forgot to mention in the comment %)).

Belco90:
Sure, but that's re-implementing the natural behavior of `React.useState` by 1) manually storing the object on ref and 2) manually triggering an update, when `useState` already handles that (at the end `useMap` is a just a shortcut for `useState` with an object and manipulation callbacks.

I would go for leaving the setState and just memoize the callbacks. As you mentioned, should be easy to solve, all current tests should pass and you can add a new one to check callbacks returned are always same instance.

xobotyi:
[...]

PR #592, opened 2019-09-05
Memoized utils methods for useMap

Granipouss:
Hey guys,

Here is my proposal following the issue https://github.com/streamich/react-use/issues/456

All feedbacks are more than welcome

Cheers

### DIFF
--- a/src/__tests__/useMap.test.ts
+++ b/src/__tests__/useMap.test.ts
@@ -106,3 +106,14 @@ it('should reset map to initial object provided', () => {
 
   expect(result.current[0]).toEqual({ foo: 'bar', a: 1 });
 });
+
+it('should memoized its utils methods', () => {
+  const { result } = setUp({ foo: 'bar', a: 1 });
+  const [, utils] = result.current;
+
+  act(() => {
+    utils.set('foo', 'baz');
+  });
+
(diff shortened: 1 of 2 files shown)


src/__tests__/useMap.test.ts
```diff
@@ -106,3 +106,14 @@ it('should reset map to initial object provided', () => {
...
+
+it('should memoized its utils methods', () => {
+  const { result } = setUp({ foo: 'bar', a: 1 });
+  const [, utils] = result.current;
+
+  act(() => {
+    utils.set('foo', 'baz');
+  });
+
+  expect(result.current[1]).toBe(utils);
```
Belco90: You are checking here that `utils` object is the same instance, but that doesn't mean that returned callbacks are same instances, so you would have to check that `utils.get` is `result.current[1].get` and so on.

xobotyi: In fact it does mean that internals are the same (due to `useMemo` instead of bunch of `useCallback`), but that, more precise check of `result.current[1].get` wont be excessive.

Granipouss: It is indeed working just fine, but I guess an additionnal test assertion can make it more visible

PR #595, opened 2019-09-06
improve useEvent type

macinjoke:
The arguments of useEvent, name, handler, options are weak type now. 

```
const useEvent = (
  name: string,
  handler?: null | undefined | ((event?: any) => void),
  target: null | UseEventTarget = defaultTarget,
  options?: any
) => 
```

But name and handler, options can be infered from target. So I improved the type of useEvent arguments.
This makes happy coding when using useEvent() name and handler, options.

Also I added a test for useEvent.ts.

### DIFF
--- a/src/useEvent.ts
+++ b/src/useEvent.ts
@@ -3,25 +3,32 @@ import { isClient } from './util';
 
 export interface ListenerType1 {
   addEventListener(name: string, handler: (event?: any) => void, ...args: any[]);
-
-  removeEventListener(name: string, handler: (event?: any) => void);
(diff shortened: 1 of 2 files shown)


src/useEvent.ts
```diff
@@ -30,11 +37,17 @@ const useEvent = (
     if (!target) {
       return;
     }
-    const fn: any = (target as ListenerType1).addEventListener || (target as ListenerType2).on;
```
macinjoke: delete `as` by using new isListenerType1 func.

macinjoke:
Can you review it?

PR #607, opened 2019-09-15
Use tslib to dedupe TypeScript helper functions from built files

tanem:
Recently noticed that TS helper functions are being inlined into the built files. Using [`tslib`](https://github.com/microsoft/tslib#tslib) seems like the simplest way to resolve this issue right now?

Example before:

```ts
// esm/useGetSetState.js
var __assign = (this && this.__assign) || function () {
    __assign = Object.assign || function(t) {
        for (var s, i = 1, n = arguments.length; i < n; i++) {
            s = arguments[i];
            for (var p in s) if (Object.prototype.hasOwnProperty.call(s, p))
                t[p] = s[p];
        }
        return t;
    };
    return __assign.apply(this, arguments);
};
import { useCallback, useRef } from 'react';
import useUpdate from './useUpdate';
var useGetSetState = function (initialState) {
...
```

Example after:

```ts
// esm/useGetSetState.js
import * as tslib_1 from "tslib";
import { useCallback, useRef } from 'react';
import useUpdate from './useUpdate';
var useGetSetState = function (initialState) {
...
```


tanem:
Anything I can do to help move this along? It's a pretty simple change so hopefully not too much trouble to review 🙏

xobotyi:
Tbh - no sense here.
if not to use tslib cjs version weights 464kb ungziped;
if to use it - 513kb

xobotyi:
@wardoost @streamich what do you think?

tanem:
> Tbh - no sense here.
> if not to use tslib cjs version weights 464kb ungziped;
> if to use it - 470kb

Ok, wonder what the numbers are like for gzipped & `esm` builds (`esm` is what prompted me to make this PR).

Seems this is the issue `tslib` was designed to help with, but over to you folks, thanks for taking the time to consider the PR 🙂

xobotyi:
@tanem as for ESM - 436k without and 445k with tslib (ungzipped again).
It is definely makes more sense on big projects with a lot more files than `react-use`.

Possibly it would be better for projects that using tslib to have it, as it has almost no differense for others.

Thus i'll merge it.

tanem:
Ok - thanks @xobotyi!

wardoost:
Thanks 👏

@xobotyi Could you squash before merging into master and use the [Angular commit types](https://github.com/angular/angular/blob/master/CONTRIBUTING.md#type) for the commit message? We do this to have a clearer project history + the change logs are based on these commit messages.

Issue #478, opened 2019-07-21
useMouse & useMouseHovered initial state is 0


thevtm:
First of all I really enjoy using this library. 
It has helped me transition to hooks really easily. :smile: 

I'm having this bug using `useMouse` that on the first render life cycle the values provided are 0 until the user moves the mouse. 
That leads to my component being rendered in the wrong position for a few ms.

I'd recommend to use `NaN` instead of `0` for initial state.
So we would be able to detect that case.

xobotyi:
@thevtm maybe better `null` or `undefined`? cuz `NaN !== NaN === true`

blnk-space:
Hi guys, I came across the same issue and I have made a PR [#610](https://github.com/streamich/react-use/pull/610) of how I solved this issue in my project.

FezVrasta:
What happened to the PR? I'd really need a `isDirty` check

PR #610, opened 2019-09-16
Flag to track if useMouseHovered is dirty

blnk-space:
A solution to #478 

We could use NaN/null/undefined etc. for the variable values (i.e. docX, docY, elX and elY) but that would translate to poor DX overall. Moreover, it would break backward compatibility.

It would be much easier for those who are concerned with making updates after the mouse has moved once, to be able to wrap all their logic in a block that checks if the mouse has moved once, whilst the original values still stay 0 and not null or something.

I have given the name "dirty" to this flag.

Let me know what you think!

### DIFF
--- a/docs/useMouse.md
+++ b/docs/useMouse.md
@@ -13,14 +13,20 @@ import {useMouse} from 'react-use';
 
 const Demo = () => {
   const ref = React.useRef(null);
-  const {docX, docY, posX, posY, elX, elY, elW, elH} = useMouse(ref);
+  const {docX, docY, posX, posY, elX, elY, elW, elH, dirty} = useMouse(ref);
 
   return (
     <div ref={ref}>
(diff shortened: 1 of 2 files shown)


blnk-space:
@thevtm does this solve your use case as well?

@xobotyi what do you think?

thevtm:
It's a nice solution.

I would change the name of the flag to something like "touched".

streamich:
Can we check `elW`, `elH` for "dirtiness"?

```ts
const {docX, docY, posX, posY, elX, elY, elW, elH, dirty} = useMouse(ref);
const isDirty = !elW;
```

Also, `docX` and `docY` could probably be set from the very beginning. Other values could possibly be set faster if we used `useLayoutEffect` hook instead of `useEffect`.

blnk-space:
@streamich yes that does make sense and works without changing the API.

PR #613, opened 2019-09-18
useQueue hook

droganov:
implements simple queue


wardoost:
Functionally is very similar to `useList`, what reasons do you think warrant a separate hook?

droganov:
@wardoost with `useList` you'll have to write extra code when you need queue

`set([...list, member])` is not `add(member)`
`last` is better than ugly `list[list.length-1]`

xobotyi:
@droganov thus youd better implement `shift` and `unshift` methods. 
Due to absense of specification which queue it is LIFO or FIFO, i think it should have both implementation

droganov:
@xobotyi LIFO is the stack, not queue, I'll add it as a different hook

xobotyi:
@droganov i know that it is named stack but still a sort of queue %) [proof](https://en.wikipedia.org/wiki/FIFO_(computing_and_electronics))

droganov:
@xiaoboost no, stack is different, if we want complicate things we can use list
I prefer it to be a separate hook

droganov:
@xobotyi yesterday it was late, let me elaborate please.

You're right that technically stack and queue are very similar, one appends, another one prepends and they share rest of the code.

But if we add both methods, we gonna get a methodical problem: if it is a stack, then why I can use it as queue? If it is not really a stack or queue then what it is? A sort of buffer?

Suppose we name it somehow, and wrote some code using our hybrid. It will be a code which is harder to maintain because you can not simply season if that a stack or queue with the first line, you'll have to read down and see how it's used.

This is why I want idiomatic functions. As for code sharing, we can solve that differently, I'll create a factory and use it to produce both queue and stack, so they will have their own write methods and share rest of the code.

wardoost:
@droganov `useList` does have a push method to add an item to the end of the list.

From the discussion with @xobotyi it's clear there will be confusion as to how a queue should grow and shrink. Using `shift`, `unshift`, `pop` and `push` would work fine for a generic list hook, maybe we should add all of those methods as now that hook only has `push` but that's beside this PR.

I agree something named a queue should be more idiomatic. The append/prepend confusion could be cleared up by having a direction/reverse parameter.

Isn't it also more consistent to use `add` and `removeLast` or `push` and `pop` but not a mix of those duos?

droganov:
@wardoost as for naming I find `add` good because it's natural and short, so I would keep it.
as for `pop` it would be more natural to name it `remove`, but `pop` is shorter.

Should I rename `pop` to `remove`?

wardoost:
yeah, `remove` makes sense if you're using `add`.

droganov:
@wardoost renamed

PR #620, opened 2019-09-24
feat: update useInterval deps

zyy7259:


wardoost:
This change would recreate the callback on every interval which we are trying to avoid with `useRef`, [this paragraph](https://overreacted.io/making-setinterval-declarative-with-react-hooks/#refs-to-the-rescue) of Dan Abramov's article explains it better than I can 😬

I have opened #626 to improve documentation and variable naming.

zyy7259:
I don't understand why `This change would recreate the callback on every interval`.

If I understand `useEffect` correctly, without the second deps array, the effect following will re-run every time when the component renders, which I don't think is necessary.

```js
  useEffect(() => {
    latestCallback.current = callback;
  });
```

With the second deps array, the effect will only re-run when the `callback` changes.

```js
  useEffect(() => {
    latestCallback.current = callback;
  }, [callback]);
```

If you scroll to the section [Just Show Me the Code](https://overreacted.io/making-setinterval-declarative-with-react-hooks/#just-show-me-the-code) of Dan's blog. The implementation do use a second argument.

```js
  // Remember the latest callback.
  useEffect(() => {
    savedCallback.current = callback;
  }, [callback]);
```

wardoost:
Generally you would be using this hook with an inline function:
```js
useInterval(() => {
  setCount(count + 1);
}, delay);
```
This inline function is a new function on every render so `useEffect` would run every time which results in the same behaviour as `useEffect` without a deps array as you pointed out.

There might be a possibility for a performance improvement by adding a 3rd parameter with a dependency array but this might break the special case of having a callback stored in a ref so it can be used with `setInterval` as described in the blog post.

I have opened a PR https://github.com/gaearon/overreacted.io/pull/518 to update the blogpost as it's inconsistent with examples in the [Refs to the Rescue!](https://overreacted.io/making-setinterval-declarative-with-react-hooks/#refs-to-the-rescue) paragraph. Let's see what comes out of that 🙂

zyy7259:
> There might be a possibility for a performance improvement by adding a 3rd parameter with a dependency array but this might break the special case of having a callback stored in a ref so it can be used with `setInterval` as described in the blog post.

```js
  // Remember the latest callback.
  useEffect(() => {
    savedCallback.current = callback;
  }, [callback]);
```

How would the code above break the special case of having a callback stored in a ref? If the callback does change, the effect will execute and the ref will point to the updated callback.

wardoost:
Your code snippet doesn't break anything but adding the callback to the dependency list of the effect is redundant because it's meant to be used with an inline function `() => { ... }`. This inline function is a new function on every render so useEffect would run every time which results in the same behaviour as useEffect without a deps array as you pointed out.

Currently we want this callback to always be refreshed with the latest state and/or props to be useful with `setInterval`. In the comment you quoted I was thinking of adding a deps array to allow for optimisation.

```js
const useInterval = (callback, delay, deps) => {
  const savedCallback = useRef(() => {});

  useEffect(() => {
    savedCallback.current = callback;
  }, deps);
  
  ...
```

This way you would able to have your callback only refresh when certain deps change.

PR #626, opened 2019-09-26
useInterval refactor and docs

wardoost:
Renamed callback ref variable and improved documentation example on how to pause the interval.

### DIFF
--- a/src/useInterval.ts
+++ b/src/useInterval.ts
@@ -1,17 +1,18 @@
 import { useEffect, useRef } from 'react';
 
 const useInterval = (callback: Function, delay?: number | null) => {
-  const latestCallback = useRef<Function>(() => {});
+  const savedCallback = useRef<Function>(() => {});
 
   useEffect(() => {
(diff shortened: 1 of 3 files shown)


src/useInterval.ts
```diff
@@ -1,17 +1,18 @@
 import { useEffect, useRef } from 'react';
 
 const useInterval = (callback: Function, delay?: number | null) => {
-  const latestCallback = useRef<Function>(() => {});
+  const savedCallback = useRef<Function>(() => {});
 
   useEffect(() => {
-    latestCallback.current = callback;
+    savedCallback.current = callback;
   });
```
zyy7259: I think this effect needs a second argument. Here is a detailed comment from [another pull request](https://github.com/streamich/react-use/pull/620#issuecomment-536245044).

```suggestion
  }, [callback]);
```

wardoost: See https://github.com/streamich/react-use/pull/626/files#issuecomment-541392191

PR #638, opened 2019-10-02
Fix useHoverDirty deps

rskull:
`useHoverDirty` is not working.


wardoost:
I did some testing and the current version actually implements it correctly. Also see this quote from https://github.com/facebook/react/issues/16121#issuecomment-511369830:

> However, it doesn’t make sense to add ref.current as a dependency. For the same reason it doesn’t make sense to add window.myVariable. When it updates, React won’t know about it, and won’t update your component.

Issue #634, opened 2019-09-28
useStateList


ralphchristianeclipse:
currently useStateList returns
```js
{
        state: stateSet[currentIndex],
        next: next,
        prev: prev,
    };

```

it would be great if you also return the currentIndex which will be used as a page indicator.

```js
{
        state: stateSet[currentIndex],
        next: next,
        prev: prev,
        index: currentIndex

    };
```

PR #640, opened 2019-10-03
feat: added currentIndex return value

wilbo:
Added the currentIndex to the return value, as requested by issue: https://github.com/streamich/react-use/issues/634


xobotyi:
@wilbo please update the readme and story and write tests.
Otherwise - LGTM.

xobotyi:
Close in favor of #757

PR #645, opened 2019-10-04
Update screenfull to v5.0.0

jakeboone02:
This updates the dependency and the code for `useFullscreen()` to be compatible with `screenfull` v5.0.0.

#598 is broken because it doesn't fix the code.

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
     "react-fast-compare": "^2.0.4",
     "react-wait": "^0.3.0",
     "resize-observer-polyfill": "^1.5.1",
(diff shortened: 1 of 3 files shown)


jakeboone02:
@wardoost I think it has something to do with `esModuleInterop: true` in `tsconfig.json`.  That enables `allowSyntheticDefaultImports`, which has the [description](https://www.typescriptlang.org/docs/handbook/compiler-options.html) "Allow default imports from modules with no default export."  I could be wrong, but that seems to be what's going on.

Issue #604, opened 2019-09-12
useIntersection


qianL93:
It will be a sensory hook for IntersectionObserver.
Description:
`(ref: HTMLElement | React.RefObject<HTMLElement>, options: IntersectionObserverInit) => IntersectionObserverEntry`

flycj:
useIntersection  Is this feature still not available?

PR #652, opened 2019-10-06
Implement useIntersection hook.

kevinnorris:
As described in https://github.com/streamich/react-use/issues/604.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@
   - [`useGeolocation`](./docs/useGeolocation.md) &mdash; tracks geo location state of user's device. [![][img-demo]](https://streamich.github.io/react-use/?path=/story/sensors-usegeolocation--demo)
   - [`useHover` and `useHoverDirty`](./docs/useHover.md) &mdash; tracks mouse hover state of some element. [![][img-demo]](https://codesandbox.io/s/zpn583rvx)
   - [`useIdle`](./docs/useIdle.md) &mdash; tracks whether user is being inactive.
-  - [`useIntersection](./docs/useIntersection.md) &mdash; tracks an HTML element's intersection. [![][img-demo]](https://streamich.github.io/react-use/?path=/story/sensors-useintersection--demo)
+  - [`useIntersection`](./docs/useIntersection.md) &mdash; tracks an HTML element's intersection. [![][img-demo]](https://streamich.github.io/react-use/?path=/story/sensors-useintersection--demo)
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -49,6 +49,7 @@
   - [`useGeolocation`](./docs/useGeolocation.md) &mdash; tracks geo location state of user's device. [![][img-demo]](https://streamich.github.io/react-use/?path=/story/sensors-usegeolocation--demo)
   - [`useHover` and `useHoverDirty`](./docs/useHover.md) &mdash; tracks mouse hover state of some element. [![][img-demo]](https://codesandbox.io/s/zpn583rvx)
   - [`useIdle`](./docs/useIdle.md) &mdash; tracks whether user is being inactive.
+  - [`useIntersection](./docs/useIntersection.md) &mdash; tracks an HTML element's intersection. [![][img-demo]](https://streamich.github.io/react-use/?path=/story/sensors-useintersection--demo)
```
wardoost: Forgot a ` after useIntersection

docs/useIntersection.md
```diff
@@ -0,0 +1,36 @@
+# `useIntersection`
+
+React sensor hook that tracks the changes in the intersection of a target element with an ancestor element or with a top-level document's viewport.
```
wardoost: Can be useful to add "...using Intersection Observer API" with a link to https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API as the hook returns a complete intersection observer entry object?

kevinnorris:
I'm not sure how to test the main functionality of the hook. As it requires elements intersecting via scrolling I assume it may be possible with by using a library like enzyme to fully render the DOM and then force scroll events. However, this approach requires adding additional dependencies.

peter-mouland:
@kevinnorris I recently created a hook very similar to this and used shopify mocks for the testing

`import { intersectionObserver } from '@shopify/jest-dom-mocks';`

Belco90:
Never done this before for `intersectionObserver` but I'd go for mocking it and just assume it will do its thing in the real world. How to check it is up to you I guess: mocking it manually or using a third party library for it (that @shopify/jest-dom-mocks that @peter-mouland mentioned looks really interesting).

src/useIntersection.ts
```diff
@@ -0,0 +1,30 @@
...
+      observer.observe(ref.current);
+
+      return () => {
+        if (ref.current) {
+          observer.disconnect();
+        }
+      };
+    }
+    return () => {};
+  }, [ref.current]);
```
wardoost: It doesn't make sense to add `ref.current` to an effect dependencies for the same reason as described in #655 

Maybe we would want to add `threshold`, `root` and `rootmargin` from options to the dependency list of this effect?

PR #653, opened 2019-10-07
Add hook to calculate scrollbar width

maxmalov:


src/useScrollbarWidth.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  return scrollbarWidth;
+};
+
+const useScrollbarWidth = () => {
+  const [width, setWidth] = useState<number | undefined>();
+
+  useEffect(() => {
+    if (isClient) {
+      setWidth(getScrollbarWidth());
```
maxmalov: Probably it makes sense to cache calculated value somewhere, since it won’t change frequently

xobotyi: There are some more stuff related to sbw.
I'll bring here my implementation of detector from [react-scrollbars-custom](https://github.com/xobotyi/react-scrollbars-custom)

PR #662, opened 2019-10-09
fix: move @types/react-wait to dependencies, closes #661

### LINKED ISSUE #661: @types/react-wait not included in dependencies

TypeScript build fails with error:
```
node_modules/react-use/lib/useWait.d.ts:1:33 - error TS7016: Could not find a declaration file for module 'react-wait'. 

'....../node_modules/react-wait/dist/react-wait.js' implicitly has an 'any' type.
  Try `npm install @types/react-wait` if it exists or add a new declaration (.d.ts) file containing `declare module 'react-wait';`

1 import { useWait, Waiter } from 'react-wait';
                                  ~~~~~~~~~~~~
```

```
echo '{}' > package.json && \
  npm i typescript react-use@latest && \
  echo "import { useWait } from 'react-use'; console.log(useWait);" && \
  npx tsc --init && \
  npx tsc --noEmit
```

Contrary to the expected some TypeScript types should be included as `dependencies`. This is the case with `react-wait` on which typings for `node_modules/react-use/lib/useWait.d.ts` depend (so it should indeed be included as dependency, not devDependency). More on that can be found in [TypeScript handbook](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#dependencies).

Simply reverting 49372ac should be enough to make things work again :smile:

**A little about versions:**  
- _OS_: macOS
- _Browser (vendor and version)_: n/a
- _React_: n/a
- _`react-use`_: 12.2.3
- _Did this worked in the previous package version?_ yes (before #644 was implemented) - 12.2.2 does not have the error

rkostrzewski:
More info on why and how in #661

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
   },
   "homepage": "https://github.com/streamich/react-use#readme",
   "dependencies": {
+    "@types/react-wait": "^0.3.0",
     "copy-to-clipboard": "^3.1.0",
     "nano-css": "^5.1.0",
     "react-fast-compare": "^2.0.4",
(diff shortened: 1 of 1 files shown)


wardoost:
Thanks for the correction 👍

PR #665, opened 2019-10-10
fix: useSize avoid crash in Safari 11

qianL93:
`window.removeEventListener` could be null for a **removed** iframe element in **Safari 11** while `window` is regarded as a reference to `iframe.contentWindow`.

### DIFF
--- a/src/useSize.tsx
+++ b/src/useSize.tsx
@@ -67,7 +67,7 @@ const useSize = (
     }
 
     return () => {
-      if (window) {
+      if (window && window.removeEventListener) {
         window.removeEventListener('resize', setSize);
       }
     };


wardoost:
I could not recreate the issue in Safari 11 but only able to test on 11.1 😬 Do you have more specifics as to when this issue happens or could you link any resources describing the issue?

qianL93:
👌I have open an issue for this, but I couldn't find more resources for this. It isn't a bug for react-use itself, we just do more checking to avoid unexpected error.

wardoost:
What's the point of doing more checking if there is no bug? Maybe I'm not understanding you correctly?

qianL93:
> What's the point of doing more checking if there is no bug? Maybe I'm not understanding you correctly?

There is an actually bug. 
 >  It isn't a bug for react-use itself

I mean this bug is due to a nonstandard Window Object Reference provided by Safari. AND, UNLUCKY, It is already reality.  😂

The point of doing more checking is to help the code always work well under this circumstance.

src/useSize.tsx
```diff
@@ -67,7 +67,7 @@ const useSize = (
     }
 
     return () => {
-      if (window) {
```
streamich: I'm not sure I understand how this bug happens. Is `<iframe>` removed from DOM before this line executes?

src/useSize.tsx
```diff
@@ -67,7 +67,7 @@ const useSize = (
...
+      if (window && window.removeEventListener) {
```
streamich: I'm not sure I understand how this bug happens. Is `<iframe>` removed from DOM before this line executes?

qianL93: yes, I believe it could happen, such as  under `contenteditable` element,  though it is rare. I am not sure if we should hold this scene. ?

streamich: Has it ever happened? Can you reproduce it?

qianL93: Here, I can show you how to reproduce it in a simple way by a  [demo](https://wes4r.csb.app/).

streamich: Can you please post a screenshot, I'm on Safari 13.

qianL93: 480p-gif

on iOS 11.4 (Simulator from Xcode)

PR #673, opened 2019-10-13
useVibrate

### LINKED ISSUE #43: useVibrate
See example: https://github.com/streamich/libreact/blob/master/docs/en/Vibrate.md

wardoost:
Implementation of a vibration effect hook that can be controlled with an `enabled` parameter. Will add tests if we like the hook signature of this proposal. Generally only works on Android phones...

Closes #43

### DIFF
--- a/src/__stories__/useVibrate.story.tsx
+++ b/src/__stories__/useVibrate.story.tsx
@@ -1,0 +1,20 @@
+import { storiesOf } from '@storybook/react';
+import * as React from 'react';
+import { useVibrate, useToggle } from '..';
+import ShowDocs from './util/ShowDocs';
+
+const Demo = () => {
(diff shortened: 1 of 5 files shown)


src/__stories__/useVibrate.story.tsx
```diff
@@ -0,0 +1,20 @@
+import { storiesOf } from '@storybook/react';
+import * as React from 'react';
+import { useVibrate, useToggle } from '..';
+import ShowDocs from './util/ShowDocs';
+
+const Demo = () => {
+  const [vibrating, toggleVibrating] = useToggle(false);
+
+  useVibrate(vibrating, [300, 100, 200, 100, 1000, 300]);
```
streamich:
```suggestion
  useVibrate(vibrating && [300, 100, 200, 100, 1000, 300]);
```

PR #674, opened 2019-10-14
add useStatesValidator

whoo3474:
Hello? 
The 'useStateValidate' that was posted on react-use a few days ago is in the form of posting 'useState' one by one.
I made a 'useStatesValidate'. (Added the letter s in the middle.)
I can manage multiple input, minLength, maxLength, required, regular expression pattern, error message etc. and use it as the user wants.
I made it because I wanted to manage the validation of many inputs in one place.
Can you check it out?
It's my first time contributing to open source so please give me a lot of feedback.
I'd really appreciate it if you could do that.
Thank you for reading. I hope you have a good day!

### DIFF
--- a/docs/useStatesValidator.md
+++ b/docs/useStatesValidator.md
@@ -1,0 +1,104 @@
+# `useStatesValidator`
+
+Create array that contain values that you want to verify.
+Defines objects within an array.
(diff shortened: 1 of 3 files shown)


xobotyi:
Not sure if it should be a part of library because it is very limited and hardcoded functionality.

[`useStateValidator`](https://github.com/streamich/react-use/blob/master/docs/useStateValidator.md) is way more versatile.

PR #676, opened 2019-10-14
feat: useMediatedState

xobotyi:
This hook will be usefull for forms with inputs and so on, now the function that processes raw input can be defined near the state, not "somwhere there, maybe inlined in jsx props".


src/__stories__/useMediatedState.story.tsx
```diff
@@ -0,0 +1,28 @@
+import { storiesOf } from '@storybook/react';
+import * as React from 'react';
+import { useMediatedState } from '../useMediatedState';
+import ShowDocs from './util/ShowDocs';
+
+const InputMediator = s => s.replace(/[\s]+/, ' ');
```
wardoost: I think the regex needs to be global and only replace double spaces?

```js
const InputMediator = s => s.replace(/\s\s+/g, ' ');
```

xobotyi: Typo%) recently been writing in php - always forget to puth the `g` flag in JS's regexes..

xobotyi:
@wardoost dont know why, but as for my ears `useMediatedState` sound more suitable here than `useStateMediator`. Second one sound like it wont affect the state, while fors one is for sure will.

src/__stories__/useMediatedState.story.tsx
wardoost: Doesn't the regex have to be `/\s\s+/g` to only replace double spaces?

I would also change `InputMediator` to `inputMediator` as it's the convention in JS.

docs/useMediatedState.md
```diff
@@ -0,0 +1,42 @@
+# `useMediatedState`
+
+A lot like the standard `useState`, but with mediation process.
+
+## Usage
+```ts
+import * as React from 'react';
+import { useMediatedState } from '../useMediatedState';
+
+const InputMediator = s => s.replace(/[\s]+/, ' ');
```
wardoost: Can you update the regex and rename `InputMediator` to `inputMediator`?

xobotyi: I've updated %) strange that you've accessed the old one.
About function name i'll forcepush now.

src/__stories__/useMediatedState.story.tsx
xobotyi: There can be not only space sharacters (4ex tab character) so have to replace them too.
Anyways not sure if it is important for the demo.

docs/useMediatedState.md
wardoost: This comment is in the mardown file of your hook, not the story :)

xobotyi: \*facepalm.jpg*

src/__stories__/useMediatedState.story.tsx
wardoost: Doesn't `\s` on its own cover all whitespace characters? I don't think you need the square brackets.

It's not super important for the demo but best to give a common example that is correct/

PR #679, opened 2019-10-15
docs: add contributing guidelines

### LINKED ISSUE #505: [Suggestion] add CONTRIBUTING.md
I recently submitted a PR and when setting up the project locally I had some issues. Particularly related to yarn, I have no previous experience using it and had some issues with duplicated React types). 

I think a simple "get started" section in a CONTRIBUTING.md file would be most helpful :)

wardoost:
Closes #505 

@streamich We might want to only _allow squash merging_ in the settings of this github repository? This will lead to a cleaner project history because we would only have 1 commit for each PR.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,60 @@
+# Contributing
+
+Thanks for being willing to contribute 🙌 If you contribute to this project, you agree to release your work under the license of this project.
+
+**Working on your first Pull Request?** You can learn how from this [First Contributions](https://github.com/firstcontributions/first-contributions)  guide.
+
+## Project setup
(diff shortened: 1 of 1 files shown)


Belco90:
Oh wow, this is fantastic 💯

Actually, the steps you mentioned in `Creating a new hook` section is what I wanted to check with Danger JS, so if you add a new hook it automatically checks there are corresponding story and test files at least (but didn't have time to finish that, I hope this week 🤞)

About only allowing squashing merges, I like that idea to improve conventional commits and semantic releases (we could even use [semantic-pull-requests](https://github.com/probot/semantic-pull-requests))

PR #680, opened 2019-10-15
feat: useSlider

### LINKED ISSUE #42: useSlider
Create `useSlider` hook, based on:

- https://github.com/streamich/libreact/blob/master/docs/en/Slider.md

wardoost:
Implementation of a slider hook taking a ref and options as parameters.

I would like some feedback on the hook signature of my proposal and also what options we would like to add. For example an option to clamp the return `pos` could be valuable?

Will add tests if we like the hook signature of this proposal.

Closes #42


streamich:
I love the demo :D

image

src/__stories__/useSlider.story.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+const Demo = () => {
+  const ref = React.useRef(null);
+  const state = useSlider(ref);
+
+  return (
+    <div>
+      <div ref={ref} style={{ position: 'relative', background: 'yellow', padding: 4 }}>
+        <p style={{ margin: 0, textAlign: 'center' }}>Slide me</p>
+        <div style={{ position: 'absolute', top: 0, left: state.pos, transform: 'scale(2)' }}>
```
streamich:
```suggestion
        <div style={{ position: 'absolute', top: 0, left: (100 * state.value) + '%', transform: 'scale(2)' }}>
```

src/useSlider.ts
```diff
@@ -0,0 +1,150 @@
...
+      return undefined;
+    }
+  }, [ref]);
+
+  useEffect(() => {
+    setState(prevState => ({
+      pos: Math.round(prevState.value * width),
+      length: width,
+    }));
+  }, [width]);
```
streamich: Maybe we should leave tracking the re-sizing up to the user. Users receives the `value`, which is normalized to `[0 ... 1]` and they can figure out what to do with it. Also, they can use `%` units in CSS with `value`, see proposed change to the story. Then we don't need to use `useMeasureDirty()`, which uses `ResizeObserver`, which is not supported by some older browsers.

src/useSlider.ts
```diff
@@ -0,0 +1,150 @@
...
+  pos: number;
+  length: number;
+}
+
+export interface Options {
+  onScrub: (value: number) => void;
+  onScrubStart: () => void;
+  onScrubStop: () => void;
+  reverse: boolean;
+  styles: boolean | CSSProperties;
```
streamich: Shall we add `vertical` option, too? Or should it be a different hook?

Issue #643, opened 2019-10-03
useNotification


haoyangnz:
Send notifications with Notification API.

xobotyi:
You are implementing it or it's a feature request?
Why don't you use issue template?

haoyangnz:
@xobotyi Oops - somehow missed that - I've edited the issue description.
It's a feature request.

haoyangnz:
I wonder if the community would be open to just merging some of the custom hook packages into one so we can all just use one package... There're literally 100+ packages right now.

wardoost:
Merging packages together just because they are hooks doesn't seem like a good enough reason to group packages in my opinion. NPM makes it super easy in install and manage multiple packages for your project. Merging multiple non-related packages together just because you want your `package.json` to look cleaner will only hide complexity and make it harder for new devs on your project to understand what's going on.

I think we have to consider this when adding hooks to `react-use` which we have been able to do so far. A `useNotification` hook could definitely be added in my opinion as it would map a standardised browser API to a hook. Do you have in mind a signature for this hook? Links to notification API related packages, not necessarily hook related, can also be helpful to get an idea where we can take this.

PR #681, opened 2019-10-15
Feature/use notification

kevinnorris:
Requested in #643.

### DIFF
--- a/src/useNotification.ts
+++ b/src/useNotification.ts
@@ -1,0 +1,30 @@
+import { useEffect, useState } from 'react';
+import * as thisModule from './useNotification';
+
+export async function requestPermission(setPermission) {
+  const permissionFromUser = await Notification.requestPermission();
+
+  setPermission(permissionFromUser);
+}
+
+const useNotification = (title: string, options: NotificationOptions): (() => void) => {
+  const [permission, setPermission] = useState(
+    Notification && Notification.permission ? Notification.permission : 'denied'
+  );
+
+  useEffect(() => {
+    if (permission !== 'denied' && permission !== 'granted') {
+      thisModule.requestPermission(setPermission);
+    }
+  }, [permission]);
+
+  if (permission === 'granted') {
+    return () => {
+      new Notification(title, options);
+    };
+  }
+
+  return () => {};
+};
+
+export default useNotification;
--- a/README.md
+++ b/README.md
@@ -74,6 +74,7 @@
   - [`useCss`](./docs/useCss.md) &mdash; dynamically adjusts CSS.
   - [`useDrop` and `useDropArea`](./docs/useDrop.md) &mdash; tracks file, link and copy-paste drops.
   - [`useFullscreen`](./docs/useFullscreen.md) &mdash; display an element or video full-screen. [![][img-demo]](https://streamich.github.io/react-use/?path=/story/ui-usefullscreen--demo)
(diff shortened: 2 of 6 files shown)


src/useNotification.ts
```diff
@@ -0,0 +1,30 @@
...
+}
+
+const useNotification = (title: string, options: NotificationOptions): (() => void) => {
+  const [permission, setPermission] = useState(
+    Notification && Notification.permission ? Notification.permission : 'denied'
+  );
+
+  useEffect(() => {
+    if (permission !== 'denied' && permission !== 'granted') {
+      thisModule.requestPermission(setPermission);
```
streamich: Interesting, why do you use `thisModule.requestPermission` instead of just `requestPermission`?

src/useNotification.ts
```diff
@@ -0,0 +1,30 @@
+import { useEffect, useState } from 'react';
+import * as thisModule from './useNotification';
```
streamich: Maybe we should not import this module in itself?

src/useNotification.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export async function requestPermission(setPermission) {
+  const permissionFromUser = await Notification.requestPermission();
+
+  setPermission(permissionFromUser);
+}
```
streamich: It could be that after the `await` the component is already unmounted. Something like this should work:

```ts
export async function requestPermission(setPermission, isMounted) {
  const permissionFromUser = await Notification.requestPermission();
  if (!isMounted()) return;
  setPermission(permissionFromUser);
}
```

src/useNotification.ts
kevinnorris: This allows me to mock and spy on `requestPermission` in the tests. If you know of another method to accomplish this I am very interested :)

src/useNotification.ts
kevinnorris: As above

xobotyi:
`Notification.requestPermission()` should already contain the prompt. there is no sense to add another one.

kevinnorris:
> Also, there is this convention that websites first ask the user if they would want to allow the site to show notifications and only then trigger the permission grant UI, like this:

@streamich seemed to think there was a reason for an additional prompt.

streamich:
@kevinnorris I was thinking about something more generic, like

```ts
useNotification({
  async doAskPermission () {
    return confirm('Would you like to see notifications?');
  }
});
```

not sure if it is good API.

@xobotyi the idea is, to prevent user from instinctively clicking "Block" in the permission dialog, many websites implement approach, where they first show user some dialog warning them that permission dialog is about to appear.

src/useNotification.ts
```diff
@@ -0,0 +1,30 @@
...
+    }
+  }, [permission]);
+
+  if (permission === 'granted') {
+    return () => {
+      new Notification(title, options);
+    };
+  }
+
+  return () => {};
```
streamich: Should we return `permission`, too?

src/useNotification.ts
```diff
@@ -0,0 +1,30 @@
...
+  );
+
+  useEffect(() => {
+    if (permission !== 'denied' && permission !== 'granted') {
+      thisModule.requestPermission(setPermission);
+    }
+  }, [permission]);
+
+  if (permission === 'granted') {
+    return () => {
```
streamich: Shall we allow users to provide custom `title` and `options` to this function?

PR #683, opened 2019-10-16
feat: useMultiStateValidator

xobotyi:
Acts much like the `useStateValidator` but this time it handles multiple states simultaneously.

### Example
Below validator will turn to true only if both values are even
```js
const [state1, setState1] = React.useState(1);
const [state2, setState2] = React.useState(1);

const [[isValid]] = useMultiStateValidator([state1, state2], s => [s.every(num => !(num % 2))]);
```

### DIFF
--- a/docs/useMultiStateValidator.md
+++ b/docs/useMultiStateValidator.md
@@ -1,0 +1,55 @@
+# `useMultiStateValidator`
+
+Each time any of given states changes - validator function is invoked.  
+
+## Usage
+```ts 
+import * as React from 'react';
+import { useMultiStateValidator } from 'react-use';
+
+const DemoStateValidator = (s: number[]) => [s.every((num: number) => !(num % 2))] as [boolean];
+const Demo = () => {
+  const [state1, setState1] = React.useState<number>(1);
+  const [state2, setState2] = React.useState<number>(1);
+  const [state3, setState3] = React.useState<number>(1);
+  const [[isValid]] = useMultiStateValidator([state1, state2, state3], DemoStateValidator);
+
+  return (
+    <div>
+      <div>Below fields will be valid if all of them is even</div>
+      <input type="number" min="1" max="10" value={state1}
+        onChange={(ev: React.ChangeEvent<HTMLInputElement>) => {
+          setState1((ev.target.value as unknown) as number);
+        }}
+      />
+      <input type="number" min="1" max="10" value={state2}
+        onChange={(ev: React.ChangeEvent<HTMLInputElement>) => {
+          setState2((ev.target.value as unknown) as number);
+        }}
+      />
+      <input type="number" min="1" max="10" value={state3}
+        onChange={(ev: React.ChangeEvent<HTMLInputElement>) => {
+          setState3((ev.target.value as unknown) as number);
+        }}
+      />
+      {isValid !== null && <span>{isValid ? 'Valid!' : 'Invalid'}</span>}
+    </div>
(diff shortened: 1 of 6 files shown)


src/__tests__/useMultiStateValidator.ts
```diff
@@ -0,0 +1,125 @@
...
+  });
+
+  it('should throw if states is not an object', () => {
+    try {
+      // @ts-ignore
+      getHook(defaultStatesValidator, 123);
+    } catch (err) {
+      expect(err).toBeDefined();
+      expect(err instanceof Error).toBe(true);
+      expect(err.message).toBe('states expected to be an object or array, got number');
```
Belco90: Better to use [toThrow](https://jestjs.io/docs/en/expect#tothrowerror) in this case, as this test could give a false negative at the moment if no error is thrown (and it's also the proper matcher for errors).

docs/useMultiStateValidator.md
```diff
@@ -0,0 +1,55 @@
...
+```ts 
+import * as React from 'react';
+import { useMultiStateValidator } from 'react-use';
+
+const DemoStateValidator = (s: number[]) => [s.every((num: number) => !(num % 2))] as [boolean];
+const Demo = () => {
+  const [state1, setState1] = React.useState<number>(1);
+  const [state2, setState2] = React.useState<number>(1);
+  const [state3, setState3] = React.useState<number>(1);
+  const [[isValid]] = useMultiStateValidator([state1, state2, state3], DemoStateValidator);
```
Belco90: Why is returning 2 arrays nested?

xobotyi: Same as for `useStateValidator` validity state is an array where 1st element is an overall validity state, and rest is arbitrary data. It is made to make the validity state same signature over any use whether you pass extra validity data or not.

I've ommitted here the revalidate function, this line could be
```typescript
const [validity, revalidate] = useMultiStateValidator([state1, state2, state3], DemoStateValidator);
```

src/__tests__/useMultiStateValidator.ts
xobotyi: Dont understand how it can give false negative >_>

Belco90: The method called doesn't throw any error --> there is no error to catch --> all the expects inside the catch are not called: the test pass (no expect called so far) but it shouldn't. But also, `toThrow` is proper matcher to call here, much simpler, readable and maintainable.

xobotyi: Ah, yes, this one will cause false-negative.
But it does not work for the hook somewhy. If to do: 
```typescript
expect(() => {
      // @ts-ignore
      getHook(defaultStatesValidator, 123);
    }).toThrowError('states expected to be an object or array, got number');
```
it will fail with 
```
Error: expect(received).toThrowError(expected)

Expected substring: "states expected to be an object or array, got number"

Received function did not throw

xobotyi: Somewhy, error been swallowed inside the `hook.result.error` and i have to rethrow it =\

docs/useMultiStateValidator.md
Belco90: Aah I didn't remember that [react-hooks-testing-library already catches errors and exposes them under `result.error`](https://react-hooks-testing-library.com/usage/advanced-hooks#errors) so it's probably related to that.

PR #693, opened 2019-10-21
Added ActionsTabStory for log

uxitten:
ActionsTabStory.tsx instead ConsoleStory.tsx


uxitten:
@streamich You're welcome

PR #695, opened 2019-10-22
add createBreakpoint

jakapatb:
create createBreakpoint()
default laptopL: 1440, laptop: 1024, tablet: 768
and can custom breakpoint
like this 
createBreakpoint({xl:1440, l:768, s:350})


src/useBreakpoint.ts
```diff
@@ -0,0 +1,24 @@
...
+            const setSideScreen = (): void => {
+                  setScreen(window.innerWidth)
+            }
+            setSideScreen()
+            window.addEventListener('resize', setSideScreen)
+            return () => {
+                  window.removeEventListener('resize', setSideScreen)
+            }
+      })
+      const sortedBreakpoints = Object.entries(breakpoints).sort((a, b) => a[1] >= b[1] ? 1 : -1)
```
streamich: Should we sort the breakpoints once and memoize them using `useMemo()`?

src/useBreakpoint.ts
```diff
@@ -0,0 +1,24 @@
+import { useEffect, useState } from 'react'
+
+function useBreakpoint(breakpoints: { [name: string]: number } = { laptopL: 1440, laptop: 1024, tablet: 768 }) {
+      const [useScreen, setScreen] = useState(0)
```
streamich: Usually hooks start with `use*`, maybe we can rename `useScreen` to `screen`.

src/useBreakpoint.ts
```diff
@@ -0,0 +1,24 @@
+import { useEffect, useState } from 'react'
+
+function useBreakpoint(breakpoints: { [name: string]: number } = { laptopL: 1440, laptop: 1024, tablet: 768 }) {
```
streamich: What do you think about a version where there would exist a factory function that creates the `useBreakpoint` hook?

```tsx
const useBreakpoint = createBreakpointHook({ laptopL: 1440, laptop: 1024, tablet: 768 });

const Demo = () => {
  const breakpoint = useBreakpoint();
};
```

xobotyi:
Not sure it should use that name. Absolutely unintuitive.
Don't know what name will be more suitable, but surely not this one.

src/useBreakpoint.ts
xobotyi: @streamich please, merge https://github.com/streamich/set-harmonic-interval/pull/4 and https://github.com/streamich/set-harmonic-interval/pull/3
Actually it is ASAP fixes, cuz now all IE support is broken.

PR #700, opened 2019-10-23
add typings for createReducer

mtinner:


src/createReducer.ts
```diff
@@ -25,11 +38,11 @@ const createReducer = (...middlewares) => {
       [reducer]
     );
 
-    const dispatchRef = useRef(
+    const dispatchRef: MutableRefObject<Dispatch<Action>> = useRef(
       composedMiddleware(
         {
           getState: () => ref.current,
-          dispatch: (...args) => dispatchRef.current(...args),
+          dispatch: (...args: [Action]) => dispatchRef.current(...args),
```
streamich: If it is just a single argument, do we need spreading here?

```suggestion
          dispatch: (action: Action) => dispatchRef.current(action),
```

src/createReducer.ts
```diff
@@ -39,7 +52,7 @@ const createReducer = (...middlewares) => {
       dispatchRef.current = composedMiddleware(
         {
           getState: () => ref.current,
-          dispatch: (...args) => dispatchRef.current(...args),
+          dispatch: (...args: [Action]) => dispatchRef.current(...args),
```
streamich: Same here.

mtinner:
@streamich @wardoost I actually prefere to not change any code in this PR, its only about typings. If you develop with typescript, its very usefull to know, what types are expected as params and as a return value, in the current solution on the master you have no typesafty, not at all.
It would be great to merge this and make the Typescript-Devs happy and TS-File a real TS-File ;). We still can investigate and optimize the running code, which is already running perfectly, in another PR.

xobotyi:
@mtinner @streamich @wardoost TBH i'd like to change entire thing.
I think it is not the proper way to introduce middlewares into the reducer hook (i mean it should not be a hook factory).

I'll bring some solution by the night.

PR #701, opened 2019-10-23
added isMount function for check component is mount yet

uxitten:
Sometimes we need to sure component is mount or no, for example when we send a request to a server and get data to set to state, before we need to check component is mount or no:

```jsx
function Home(){
  ...

  const [todo,setTodo] = useState([]);

  useMount(isMount => {
    async function loadTodo() {
      const result = await services.todo.list();
      if(isMount()){
          setTodo(result)
      }
    }

    loadTodo();
  });

  ...
}
```

### DIFF
--- a/src/useMount.ts
+++ b/src/useMount.ts
@@ -1,8 +1,15 @@
 import useEffectOnce from './useEffectOnce';
+import useUnmount from './useUnmount'
 
-const useMount = (fn: () => void) => {
+const useMount = (fn: (isMount?: () => boolean) => void) => {
+  let isMount: boolean = true;
+
   useEffectOnce(() => {
-    fn();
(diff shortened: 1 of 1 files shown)


xobotyi:
it is already implemented in [useMountedState](https://github.com/streamich/react-use/blob/master/docs/useMountedState.md) hook;

uxitten:
@xobotyi Thank you, I didn't see this, but I think my proposal can be another way for checking component is mount yet

xobotyi:
The function you pass to `useMount` is called once the component is mounted, there is no sense in checking the mounted state, because it already is.

uxitten:
I agree with you 🙏

xobotyi:
Sure =)

PR #706, opened 2019-10-25
feat: add `useLast` hook

garrettmaring:


src/useLast.ts
```diff
@@ -0,0 +1,15 @@
+import { useEffect, useRef } from 'react';
+
+const useLast = <T>(state: T, predicate: (state: T) => boolean): T | undefined => {
```
streamich: Shall we instead extend `usePrevious` hook and add an optional `predicate?` paramter?

```ts
const usePrevious = <T>(state: T, predicate: (state: T) => boolean = () => true): T | undefined => {
```

garrettmaring: Sounds good to me 👍I wasn't sure if we'd want to keep things separate or not.

I'll update the PR to enhance `usePrevious`.

garrettmaring: Looking at this a bit more, there already is a `usePreviousDistinct` hook. That hook provides a `compare` argument which can be used to implement any predicate. The naming of that hook implies that it has to do with "distinct" (non-equal) values but give that one can provide any predicate, it's a bit more generic than that.

Issue #708, opened 2019-10-25
Remove the react-wait from package?


xobotyi:
Do we really need that reexport in library?
It's only reexporting the default hook provided by `react-wait` itself. As for me there is absolutely no sense to have that dependency.

qianL93:
Agreed, It is unnecessary to export again. And we will have React.Suspense for most  scenarios.

xobotyi:
@streamich maybe merge #711 and do a major release?

streamich:
@xobotyi I've sent you a Slack invite.

PR #711, opened 2019-10-27
feat(useWait): remove useWait hook

xobotyi:
As an implementation of #708


streamich:
Shall we make it a breaking change? I.e. delay merging until we release next major.

xobotyi:
@streamich i'm pretty sure that all react-wait users are using directly react-wait's hook.
I think we can make a **patch** version with a deprecation message and **minor** version with deleted hook, IMO no sense to wait untill the major (actually i want to do some library sources sanitization before v13 release).

xobotyi:
Also `react-wait` is abandoned already for a year and it's types defenition beaks most of current react version upgrades. #622 #577 #686 and soo on strictly related to react and typescript

PR #714, opened 2019-10-27
feat(build): tweaked builds and added esnext build

xobotyi:
Now all builds are condensed in one `/dist` directory so there is rootdir no more garbaging due to development stage.
Furthermore lang level for CJS and ESM build decreased to ES3 to support the maximum amount of browsers.

Added esnext build: untraspiled ESM version.


tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es5",
+    "target": "es3",
```
streamich: Why this change?

xobotyi: For wider ie support, for those who does not transpile `node_modules`

streamich: I'm not sure we need to support ES3 builds. Do we? If somebody wants to downscale all the way down to ES3 it they should do it themselves. But if we do want to provide ES3 build, we would need to keep `/lib` folder and create separate `/es3` or `/dist/es3` folder.

xobotyi:
There is no convention about es3/5 tbh, main should point to es3/5 with cjs module resolution, modules should point to es3/5 esm resolution and esnext should point untranspiled esnext syntax if library has any.

xobotyi:
I'm not ensist on es3, I'm okay with es5, but pretty confident that cjs version should be stored in dist/cjs, cuz dist/lib giving no clue what is inside, while cjs have straight tip.

xobotyi:
If to watch react's [package.json](https://github.com/facebook/react/blob/master/packages/react/package.json) thay have cjs version in respective directory.
I think we can make a breaking change.

As from my experience - lib directory used mostly in libraries distributing the single version. All others has more obvious directories naming. At least it's been transformed to lib/cjs and so on.

streamich:
The idea is that the `/lib` folder contains the "main" (ES5) version. Which you can use to import hooks directly. It should be easy to import individual hooks, currently it is

```ts
import useAudio from 'react-use/lib/useAudio';
```

which works well for my taste. It could be simplified to

```ts
import useAudio from 'react-use/useAudio';
```

and then `dist` folder could contain everything else.

PR #715, opened 2019-10-27
feat(storybook): moved stories to separate root dir & updated config

xobotyi:
feat(storybook): moved all stories to separate root dir /stories;
feat(storybook): updated webpack config to export a function to achieve full-control mode, instead of deprecated Extend Mode;


xobotyi:
@streamich may i merge?

PR #776, opened 2019-11-15
Modify useTitle.ts' logic

zsh2401:
Restore the previous title when the current component is destroyed.

### DIFF
--- a/src/useTitle.ts
+++ b/src/useTitle.ts
@@ -1,14 +1,22 @@
-import { useRef } from 'react';
-
-function useTitle(title: string) {
-    let prevTitleContainer = useRef<string>();
-    prevTitleContainer.current = document.title;
-    document.title = title;
-    useEffect(()=>{
-        return ()=>{
-            document.title = <string>prevTitleContainer.current;
-        }
-    });
+import { useRef, useEffect } from 'react';
+export interface UseTitleOptions {
+  restoreOnUnmount?: boolean;
+}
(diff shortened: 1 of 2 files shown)


src/useTitle.ts
```diff
@@ -1,11 +1,14 @@
 import { useRef } from 'react';
 
 function useTitle(title: string) {
```
FlickerLogicalStack: I think upgraded `useTitle` can be like:

```suggestion
import { useEffect, useRef } from 'react';

type Options = {
  restoreOnUnmount?: boolean;
};

function useTitle(
  title: string,
  { restoreOnUnmount = false }: Options = {}
) {
  const prevTitle = useRef(document.title);

  document.title = title;

  useEffect(() => {
    if (restoreOnUnmount) {
      return () => {
        document.title = prevTitle.current;
      };
    }
  }, []);
}

```

The main point in fact that restoration must be optional

streamich: @zsh2401 can we make title restoration on un-mount optional, like @FlickerLogicalStack proposed, so this change is not breaking?

zsh2401: @streamich I DO AGREE! I will do it as soon as possible!

zsh2401:
@streamich  ALL DONE! I have modified code so that it could restore previous title optionally.I have also renamed 'useTitle.test.tsx' to 'useTitle.test.ts', because it's looks like not right somehow.

Issue #785, opened 2019-11-20
useLocalStorage not updating if many components watch the same key


TylerR909:
Note: I'm already working on a PR for this.

**What is the current behavior?**
Multiple calls to `useLocalStorage('myKey')` from different components are not staying in sync as the value changes.

**Steps to reproduce it and if possible a minimal demo of the problem.**

Seems that all the Sandboxes are failing to install `react-use` right now but slap this somewhere and the problem will become apparent:

```js
import useLocalStorage from 'react-use/lib/useLocalStorage';

const Increment = () => {
  const [count, setCount] = useLocalStorage('count', 0);
  return (<h1 onClick={() => setCount(count+1)}>Increment</h1>);
}

const Decrement = () => {
  const [count, setCount] = useLocalStorage('count', 0);
  return (<h1 onClick={() => setCount(count-1)}>Decrement</h1>);
}

const App = () => (
  <>
    <Increment />
    <Decrement />
  </>
)
```

Basically it's _initializing_ from localStorage into individual instances of `useState` and then never reading from localStorage again, so if it's changed out from under you... (1) you never see the update and (2) if _you_ update it you clobber whatever was there before.

**What is the expected behavior?**

If my localStorage value changes, all my components watching that value pick up the change.

nicgirault:
Any news on this issue? The problem is still here. Could I help making this fix released?

TylerR909:
Looks like it got merged into v14.0 which isn't on `master` yet. But is also `75 commits ahead, 121 commits behind master.`

Also the unit test that specifically checks against this ticket is commented out. https://github.com/streamich/react-use/blob/v14.0/tests/useLocalStorage.test.ts#L86

gimbo:
Dang. I was hoping to replace a use of https://github.com/astoilkov/use-local-storage-state with this. Guess I won't be doing that.

xobotyi:
If anyone interested it is implemented in upconig hook of `@react-hookz/web` react-hookz/web#43

gimbo:
[...]

PR #786, opened 2019-11-20
Fix use local storage

TylerR909:
Addresses #785 

# Description  
Basically a total rewrite of `useLocalStorage` to ensure updates to the keyed value are picked up by all components watching for it.

## Type of change

I don't _believe_ this will cause any breaking changes. But I did end up tearing out almost all the old code and fundamentally changed how this hook works. 🤷‍♂ 

# Checklist
  - [ ] better commit message
  - [ ] squash?
- [x] Add documentation 
  - no changes
  - no changes
  - Could use more input for proper types, like `raw=true implies value=string`


TylerR909:
Hihi,

Although I only filed the one bug (#785) I was also running into other issues while trying to use this, like it over-serializing values (`"foo"` ==> `'"\\"foo\\""'`) and some funkiness around initial state.

Running my test suite over the original hook gave these results:

image

Not sure why so many of the earlier ones are ❌, perhaps as a consequence of `useEffect` and many of those tests not rerendering (just checking the initial value). Those two :x: in the middle though are the catalyst for this PR - Over-stringifying and hooks falling out of sync.

I'm more than happy to take a couple days to work with you to tighten the bolts on this one and get it ready to go! Let me know what else you'd like to see from me.

I also previously handled #557 so I'm a little oversensitive to the `react-hooks/rules-of-hooks` eslint rules, thus the tests.

src/useLocalStorage.ts
```diff
@@ -1,40 +1,52 @@
...
+        /**
+         * If user is in private mode or has storage restriction
+         * localStorage can throw. Also JSON.stringify can throw.
+         */
+      }
+    },
+    [state, raw]
+  );
+
+  useEffect((): void => {
```
TylerR909: I could probably swap this out for `useEffectOnce`

ayush987goyal:
Hi @TylerR909 ,

Could you please rebase your branch from master. It seems like there have been more changes to the `useLocalStorage` hook. Maybe you can make changes on top of it?

Thanks

mayteio:
Just ran into #785 and unfortunately have to find another implementation until this is merged. Be great to get it in there - let me know if I can help.

TylerR909:
I'll try to check this out in the next couple days and get the PR updated. Need to remind myself what I was doing when I opened it 2 months ago 🤷‍♂

src/useLocalStorage.ts
```diff
@@ -1,40 +1,54 @@
...
-type SetStateAction<S> = S | ((prevState: S) => S);
+import { useMemo, useCallback, Dispatch, SetStateAction } from 'react';
+import useEffectOnce from './useEffectOnce';
 
 const useLocalStorage = <T>(key: string, initialValue?: T, raw?: boolean): [T, Dispatch<SetStateAction<T>>] => {
-  if (!isClient) {
+  if (!isClient || !localStorage) {
     return [initialValue as T, () => {}];
   }
+  if (!key && (key as any) !== 0) {
```
roninjin10: Can it be a boolean false?

src/useLocalStorage.ts
```diff
@@ -1,40 +1,54 @@
...
-    } catch {
-      // If user is in private mode or has storage restriction
-      // localStorage can throw. JSON.parse and JSON.stringify
-      // can throw, too.
-      return initialValue;
-    }
-  });
+  let localStorageValue: string | null = null;
+  try {
+    localStorageValue = localStorage.getItem(key);
```
roninjin10: Isn't this expensive to be reading local storage every render?

roninjin10:
IMO because reading from local storage is expensive, we should not be reading it on every rerender.  I think since we want to keep it in sync we should cache the keys in an redux-like object shared by all instances of the hook rather than local state.

src/useLocalStorage.ts
TylerR909: Not really. If you're trying to hit 60fps that gives you 16ms between rerenders for all your code to run.

From various benchmarks I'm seeing, a single read takes 0.0004-0.0010ms. Completely negligible.

If reading localStorage is someone's bottleneck, they have other issues they need to figure out.

src/useLocalStorage.ts
TylerR909: Sure, I can protect against that.

TylerR909:
Most of the way through updating this locally. I had added a bunch of tests cases back in November and some of the new Serializer stuff is missing the marks (feeding me back [object Object] instead of just... the string) and I've already spent a couple hours on a Saturday trying to resolve merge conflicts.

I'll circle back on this tomorrow and try to finish updating it.

roninjin10:
@TylerR909 One test case you can add that I think your diff fixes is that setting local storage works if it is called after unmount.
```

it('should properly update the localStorageOnChange when component unmounts', () => {
  const key = 'some_key';
  const updatedValue = { b: 'a' };
  const expectedValue = '{"b":"a"}';

  const { result, unmount } = renderHook(() => useLocalStorage(key));

  unmount();

  act(() => {
    result.current[1](updatedValue);
  });

  expect(localStorage.__STORE__[key]).toBe(expectedValue);
});
```

src/useLocalStorage.ts
roninjin10: Not sure what I was asking here looking back.

src/useLocalStorage.ts
roninjin10: Ah interesting.   I've seen multiple tutorials making a hook like this that cite reading local storage as being synchronous and expensive as the reason for wanting to make a hook.  I assumed that was based on reality but it must have just been used as a contrived example.

I started on a PR that used an object store to keep all the hooks in sync while caching the data but if it's not actually an issue that is over-optimization

TylerR909:
Okie doke. Think this is mostly updated for now.

TylerR909:
> The exception is one test, which arguably is the crux of this PR, to test that two hooks stay in sync as one updates. The reason is, because I could not understand the sync logic. As I understood it, the useLocalStorage hooks would somehow re-render and every time pick the latest value from localStorage—that would keep them in sync. I'm not sure that would work in general case, I would expect that useLocalStorage use some pub/sub or event emitter mechanism under the hood to notify all instances that some key has changed.

My concern is that someone is going to have the bright idea to come in here and "Optimize" useLocalStorage, all the tests are going to pass, and we're going to see a regression for something we _specifically_ fixed in this PR. That is the test that is supposed to prevent this.

I disagree with commenting that test out. Even if someone changes the implementation to pub/sub, that test _as written_ should still pass.

> I'm not sure that would work in general case

Based on what? I did some research when someone else brought up the same concern and was seeing localStorage reads take 0.0004-0.0010ms which is basically negligible if you're targeting 60fps. If you have 20,000 components reading localStorage and it's going slow, your problem is not the speed of localStorage.

> I could not understand the sync logic

Yeah, not my best code in the world. Basically it used to use `const [state, setState] = useState(...);`, but now it manually calculates `const state = useMemo(...)` and `const setState = useCallback(...)` with a `useEffectOnce` to _call_ `setState` if the value in localStorage has never been set before.

Once you break it down into `calc state` and `calc setState` it's a little easier to reason around.

I don't recall what I was trying to use this hook on back in November when I made these updates but it rendered the hook _entirely_ useless to me because one component was setting localStorage when a query string changed (or something) and some View component absolutely refused to see the change until remounting.

So yes this is an important test to include. Even if the implementation changes.

PR #788, opened 2019-11-20
fix: useAsyncFn does not discard old promises and might produce races

dubzzz:
# Description

The current implementation of `useAsyncFn` can cause race conditions between two Promises whenever the callback is called twice in a row as it never discard previous queries. The latest query is considered to be the results.

As consequence, we can leverage this bug in `useAsync` with this scenario:

- `<UserProfile userId={1} />` rendered - initial render
- `<UserProfile userId={2} />` rendered - props updated
- Query for userId=2 succeeds - we see the profile of 2
- Query for userId=1 succeeds - we replace the profile of 2 by the one of 1 while userId=2

```
function UserProfile({ userId }) {
  const {value} = useAsync(async () => await fetchUserInfo(userId));
  return <>{value}</>;
}
```

## Type of change

# Checklist
- [x] Add documentation

### DIFF
--- a/src/useAsyncFn.ts
+++ b/src/useAsyncFn.ts
@@ -1,4 +1,4 @@
-import { DependencyList, useCallback, useState } from 'react';
+import { DependencyList, useCallback, useState, useRef } from 'react';
 import useMountedState from './useMountedState';
 
 export type AsyncState<T> =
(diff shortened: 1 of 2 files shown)


dubzzz:
I cherry-picked the fix I did in #787 to ensure that the tests of `useAsyncFn` were still green

PR #800, opened 2019-11-25
Modify createBreakpoint.ts logic

ppeeou:
# Description

logic change

# Checklist
- [] Comment the code, particularly in hard-to-understand areas
- [ ] Add documentation

### DIFF
--- a/src/createBreakpoint.ts
+++ b/src/createBreakpoint.ts
@@ -1,7 +1,26 @@
 import { useEffect, useState, useMemo } from 'react';
 
+interface Screen {
+  laptopL: number;
+  laptop: number;
+  tablet: number;
+}
+
+enum ScreenSize {
+  LAPTOPL = 1440,
+  LAPTOP = 1024,
+  TABLET = 768,
+}
+
+const NAME = 0;
+const WIDTH = 1;
+
 const createBreakpoint = (
-  breakpoints: { [name: string]: number } = { laptopL: 1440, laptop: 1024, tablet: 768 }
+  breakpoints: Screen = {
+    laptopL: ScreenSize.LAPTOPL,
+    laptop: ScreenSize.LAPTOP,
+    tablet: ScreenSize.TABLET,
+  }
 ) => () => {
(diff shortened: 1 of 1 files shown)


src/createBreakpoint.ts
```diff
@@ -1,7 +1,26 @@
...
+  LAPTOPL = 1440,
+  LAPTOP = 1024,
+  TABLET = 768,
+}
+
+const NAME = 0;
+const WIDTH = 1;
+
 const createBreakpoint = (
-  breakpoints: { [name: string]: number } = { laptopL: 1440, laptop: 1024, tablet: 768 }
```
streamich: The idea of `{ [name: string]: number }` is that user can specify any number of breakpoints. I don't think we want to limit it to only `laptopL`, `laptop` and `tablet`.

ppeeou: right! I didn't understand all your thoughts. 
that part reverted :)

Can you see pr?
I'll try to look into this repo more.

src/createBreakpoint.ts
```diff
@@ -15,17 +17,14 @@ const createBreakpoint = (
...
-    if (screen >= width) {
-      return name;
-    } else {
-      return acc;
-    }
-  }, sortedBreakpoints[0][0]);
-  return result;
+
+  const result =
+    sortedBreakpoints.find(([_, width]) => width >= screen) || sortedBreakpoints[sortedBreakpoints.length - 1];
```
streamich: I guess the main change here is `width >= screen` vs `screen >= width`. What is the advantage of the proposed approach?

This would be a breaking change. I think we can do this change without it being breaking. Possibly,`createBreakpoint` function could receive some parameter that would change its behaviour.

ppeeou: huh... it's my fault
`createBreakpoint.ts` This file does not currently exist in the test case.
Can I make this and put it up together?

streamich: If you could add tests for this file and make it support, both, `>=` and `<` (or possibly user defined function) would be great.

PR #801, opened 2019-11-25
fix useIntersection cleanup

mtinner:
# Description

useIntersection did not cleanup. Fixed buggy behavior and memory issue.

## Type of change


streamich:
I don't understand. What difference does it make when it is wrapped in the `intersect` function? I can't see any other difference.

PR #802, opened 2019-11-25
Add createbreakpoint test case

ppeeou:
# Description
 add createBreakpoint.ts testcase

[ref](https://github.com/streamich/react-use/pull/800#discussion_r350097923)


ppeeou:
right..!  I just changed it as you say

ppeeou:
@malithmcr  Can you check it?

PR #821, opened 2019-12-05
fix: useUpdate hitting maxInt, failing to trigger rerender

o-alexandrov:
# Description

useUpdate hook can hit maxInt, failing to trigger rerender as the value we pass must always be different

## Type of change

# Checklist
- [ ] Add documentation

### DIFF
--- a/src/useUpdate.ts
+++ b/src/useUpdate.ts
@@ -1,16 +1,9 @@
 import { useCallback, useState } from 'react';
 
-/**
- * MIN & MAX safe integers are literals due to no support in IE
- */
-const minInt = -1000000000
-const maxInt = 9007199254740991
-const incrementParameter = (num: number): number => {
-  return num !== maxInt ? (num += 1) : minInt;
-}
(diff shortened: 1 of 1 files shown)


o-alexandrov:
@streamich to be honest, I haven't faced this issue myself, but when I read this code in `react-use`, I thought to propose it just to prevent a possibility, since one value is used across the whole app and we use `useUpdate` everywhere

streamich:
If you have 1,000 components on your site and each of them re-renders 60 times per second, your site would need to run for 5,000 years to hit max integer boundary (if I did the math correctly).

```
9007199254740991 / 1000 / 60 / 60 / 60 / 24 / 365 = 4760
```

o-alexandrov:
@streamich your math is correct:
```
maxInt / 1000 components / 60 frames in a sec / 60 seconds in a min / 60 mins in an hour / 24 h in a day / 365 days
```

After you mentioned it, I realized it'd never hit this large number in normal use 😄 
Maybe in web games there is a potential of this to happen, but I now think we shouldn't cover such rare case.

PR #825, opened 2019-12-08
feat: useScrollbarWidth hook;

xobotyi:
# Description

This hook uses [@xobotyi/scrollbar-width](https://github.com/xobotyi/scrollbar-width) underneath, which is well tested and grown from my scrollbars component.

This PR absorbs #653 

## Type of change

# Checklist
- [x] Add documentation

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,7 +79,6 @@
     "@testing-library/react-hooks": "3.2.1",
     "@types/jest": "24.0.23",
     "@types/react": "16.9.11",
-    "@xobotyi/scrollbar-width": "1.5.0",
     "babel-core": "6.26.3",
     "babel-loader": "8.0.6",
     "babel-plugin-dynamic-import-node": "2.3.0",
--- a/package.json
+++ b/package.json
@@ -78,6 +78,7 @@
     "@testing-library/react-hooks": "3.2.1",
     "@types/jest": "24.0.23",
     "@types/react": "16.9.11",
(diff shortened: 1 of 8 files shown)


xobotyi:
@streamich the only thing im not sure in - should it be exported by default and be a default dependency or make it optional?
@xobotyi/scrollbar-width is [tiny enough](https://bundlephobia.com/result?p=@xobotyi/scrollbar-width@1.4.1) so can be exported as default, but it is on you to decide.

streamich:
@xobotyi As I understand it can return `undefined` only when called before DOM is ready. What would be a case when React renders and DOM is not ready yet?

xobotyi:
@streamich if developer added react bundle before the closing `body` tag and/or placed in iframe.
Had several issues related to that behavior: [#115](https://github.com/xobotyi/react-scrollbars-custom/issues/115) and [#123](https://github.com/xobotyi/react-scrollbars-custom/issues/123)

package.json
```diff
@@ -78,6 +79,7 @@
     "@testing-library/react-hooks": "3.2.1",
     "@types/jest": "24.0.23",
     "@types/react": "16.9.11",
+    "@xobotyi/scrollbar-width": "1.5.0",
```
streamich: We already have it in `"dependencies"`, probably this line is not needed.

xobotyi: Forgot to remove from dev-deps.

Issue #820, opened 2019-12-05
useMediaDevices(): TypeError: Cannot read property 'addEventListener' of undefined


tayay:
I use `useMediaDevices()`.
An error occurs when `navigator.mediaDevices` is undefined.
Is it possible to prevent an error even if `navigator.mediaDevices` is undefined?
thanks.

![WS000000](https://user-images.githubusercontent.com/36296084/70218318-20b4d580-1786-11ea-84d6-f01f0444f5e1.JPG)

OBe95:
Hi @streamich, I just created a [PR](https://github.com/streamich/react-use/pull/830) for this issue

PR #830, opened 2019-12-11
fix: handle undefined mediaDevices

OBe95:
# Description

Fix: [useMediaDevices(): TypeError: Cannot read property 'addEventListener' of undefined](https://github.com/streamich/react-use/issues/820)

## Type of change

# Checklist
- [-] Comment the code, particularly in hard-to-understand areas
- [-] Add documentation
- [-] Add hook's story at Storybook
- [-] Cover changes with tests
- [-] Provide 100% tests coverage
- [-] Make sure types are fine (`yarn lint:types`).

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -1,5 +1,5 @@
 export const isClient = typeof window === 'object';
 
-export const on = (obj: any, ...args: any[]) => obj && obj.addEventListener(...args);
+export const on = (obj: any, ...args: any[]) => obj.addEventListener(...args);
 
-export const off = (obj: any, ...args: any[]) => obj && obj.removeEventListener(...args);
+export const off = (obj: any, ...args: any[]) => obj.removeEventListener(...args);
--- a/src/util.ts
+++ b/src/util.ts
@@ -1,23 +1,5 @@
 export const isClient = typeof window === 'object';
 
-export const on = (obj: any, ...args: any[]) => {
-  try {
-    obj.addEventListener(...args);
-  } catch (error) {
-    handleError(error);
-  }
-};
+export const on = (obj: any, ...args: any[]) => obj && obj.addEventListener(...args);
 
-export const off = (obj: any, ...args: any[]) => {
-  try {
-    obj.removeEventListener(...args);
-  } catch (error) {
-    handleError(error);
-  }
-};
-
-const handleError = (error: any) => {
-  if (process.env.NODE_ENV !== 'production') {
-    console.error(error);
-  }
-};
+export const off = (obj: any, ...args: any[]) => obj && obj.removeEventListener(...args);
--- a/src/util.ts
+++ b/src/util.ts
@@ -1,5 +1,23 @@
 export const isClient = typeof window === 'object';
 
-export const on = (obj: any, ...args: any[]) => obj.addEventListener(...args);
+export const on = (obj: any, ...args: any[]) => {
+  try {
+    obj.addEventListener(...args);
+  } catch (error) {
+    handleError(error);
+  }
+};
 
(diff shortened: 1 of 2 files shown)


src/util.ts
```diff
@@ -1,5 +1,23 @@
...
+  } catch (error) {
+    handleError(error);
+  }
+};
 
-export const off = (obj: any, ...args: any[]) => obj.removeEventListener(...args);
+export const off = (obj: any, ...args: any[]) => {
+  try {
+    obj.removeEventListener(...args);
+  } catch (error) {
```
streamich: I'm not sure we should be swallowing these errors. Also, these utilities are use by other hooks, too, not only by the `useMediaDevices` hook, if we want to swallow the errors, would be better to do it in the `useMediaDevice` hook directly.

OBe95: Well, I don't see why `useMediaDevices` should have a different behavior, especially, if the source of the issue is the same.
As we could have similar issues for another `addEventListener` or `removeEventListener`, I thought we'd rather handle it globally and swallow the errors in PROD env (unless I misunderstood the issue) ?

streamich: I don't think we should be swallowing these errors. You are swallowing the errors in all environments. In PROD you are not logging.

OBe95: Ah! I see you point (I interpreted swallowing as not logging).
I clearly misunderstood what was expected in the issue, how would you approach it ?
Doing it only for the `useMediaDevices` hook or just checking if the `obj` is defined before manipulating the event listeners ?

streamich: Yes, checking if object exists before manipulating it sounds good.

OBe95: Done in a new commit (don't know what you are following in the project, you rather update the commit or add new commit for feedback ?)

src/util.ts
```diff
@@ -1,5 +1,5 @@
 export const isClient = typeof window === 'object';
 
-export const on = (obj: any, ...args: any[]) => obj.addEventListener(...args);
+export const on = (obj: any, ...args: any[]) => obj && obj.addEventListener(...args);
```
xobotyi: Not sure if these functions should be touched at all. With your changes in some cases handler bind will silently fail and evelopment can be hardened. Better to have error in that case.

You should leave only changes in `src/useMediaDevices.ts`, update the PR and ill merge it.

xobotyi:
@streamich maybe make a breaking change and change the returning signature?
I dont see any reason to return object with a single property.

src/util.ts
xobotyi: As you forbid to make changes in your fork repo it should be like so:

**`useMediaDevices`:**
```typescript
import { useEffect, useState } from 'react';
import { noop, off, on } from './util';

const useMediaDevices = () => {
  const [state, setState] = useState({});

  useEffect(() => {
    if (!navigator.mediaDevices) return;

    let mounted = true;

    const onChange = () => {
      navigator.mediaDevices
        .enumerateDevices()
        .then(devices => {
          mounted &&
            setState({
              devices: devices.map(({ deviceId, groupId, kind, label }) => ({ deviceId, groupId, kind, label })),
            });
        })
        .catch(noop);
    };

    on(navigator.mediaDevices, 'devicechange', onChange);
    onChange();

    return () => {
      mounted = false;
      off(navigator.mediaDevices, 'devicechange', onChange);
    };
  }, []);

  return state;
};

export default useMediaDevices;
```

**`util`**
```typescript
export const isClient = typeof window === 'object';

export const on = (obj: any, ...args: any[]) => obj.addEventListener(...args);

export const off = (obj: any, ...args: any[]) => obj.removeEventListener(...args);

export const noop = () => {};
```

PR #831, opened 2019-12-12
refactor(useClickAway): useCallback instead of useRef & useEffect

suyingtao:
# Description

I think the code should be more simpler. 
The following two pieces of code have the same effect：

```js
  const savedCallback = useRef(onClickAway);
  useEffect(() => {
    savedCallback.current = onClickAway;
  }, [onClickAway]);
```

```js
  const savedCallback = useCallback(onClickAway, [onClickAway]);
```

## Type of change

# Checklist
- [ ] Add documentation


xobotyi:
It won't work when callback change

suyingtao:
> It won't work when callback change

that's right 👍

PR #837, opened 2019-12-15
change useUpdate hook to use `reducer` instead of `state + callback`

MHA15:
if we using reducer instead of useState then we don't need a useCallback and it's performance will increase

### DIFF
--- a/src/useUpdate.ts
+++ b/src/useUpdate.ts
@@ -1,4 +1,4 @@
-import { useCallback, useState } from 'react';
+import { useReducer } from 'react';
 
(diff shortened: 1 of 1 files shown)


streamich:
> ... and it's performance will increase

Why do you think performance will increase?

MHA15:
> > ... and it's performance will increase
> 
> Why do you think performance will increase?

no need to use `useCallback` and have no inline function

MHA15:
@streamich 
Don't you agree that this implementation is better than old one (epsilon better) ?

PR #850, opened 2019-12-26
feat: add global state hook

MrHuangJser:
# Description

## Type of change

# Checklist
- [x] Add documentation


src/useGlobalState.ts
```diff
@@ -0,0 +1,31 @@
...
+      store.state = state;
+      store.setters.forEach(setter => setter(store.state));
+    },
+    setters: [],
+  };
+
+  return (): [S | undefined, (state: S) => void] => {
+    const [globalState, stateSetter] = useState<S | undefined>(store.state);
+
+    if (!store.setters.includes(stateSetter)) {
```
streamich: Seems this `IfStatement` will be called on every re-render. Probably better to put it inside `useLayoutEffect`?

MrHuangJser: yeah, i'll pull a new request

docs/createGlobalState.md
```diff
@@ -0,0 +1,32 @@
+# `useGlobalState`
+
+A React hook witch create a globally shared state.
```
streamich:
```suggestion
A React hook which creates a globally shared state.
```

mbest:
This PR moved the following code

```ts
      if (!store.setters.includes(stateSetter)) {
        store.setters.push(stateSetter);
      }
```

to be within a `useLayoutEffect`. Are you sure that `useLayoutEffect` is synchronous? If not, you could have a race condition where the value is changed after you read the value but before you subscribe to changes.

MrHuangJser:
> In this PR, you moved the following code
> 
> ```ts
>       if (!store.setters.includes(stateSetter)) {
>         store.setters.push(stateSetter);
>       }
> ```
> 
> to be within a `useLayoutEffect`. Are you sure that `useLayoutEffect` is synchronous? If not, you could have a race condition where the value is changed after you read the value but before you subscribe to changes.

thk  your replay i'll fix improve it in next pr :)

PR #857, opened 2020-01-03
fix: generic type on event arg in onClickAway callback on useClickAway

### LINKED ISSUE #393: useClickAway: callback event definition KeyboardEvent is not enough
The TypeScript declaration for the [callback](https://github.com/streamich/react-use/blob/v9.4.1/src/useClickAway.ts#L8) is:

```TypeScript
onClickAway: (event: KeyboardEvent) => void
```

KeyboardEvent is not enough, it should be: `KeyboardEvent | TouchEvent | MouseEvent | ...` or more probably `UIEvent` and then the user should cast it to MouseEvent, KeyboardEvent... depending on the parameters he passed to useClickAway()

frfroes:
Fixes #393

# Description

I'm using generic typing on the hook definition to provide the user the ability to type the `onClickAway` callback's arg in the way it better suits her/his needs. The generic is extending and falling back to the `Event` type because it was the most generic one I could think would be relevant for this scenario. 

## Type of change

# Checklist
- [x] Add documentation


src/useClickAway.ts
```diff
@@ -3,9 +3,9 @@ import { off, on } from './util';
 
 const defaultEvents = ['mousedown', 'touchstart'];
 
-const useClickAway = (
+const useClickAway = <T extends Event>(
```
streamich: Would default help here?

```suggestion
const useClickAway = <T extends Event = Event>(
```

src/useClickAway.ts
```diff
@@ -3,9 +3,9 @@ import { off, on } from './util';
...
   ref: RefObject<HTMLElement | null>,
-  onClickAway: (event: KeyboardEvent) => void,
+  onClickAway: (event: T | Event) => void,
```
streamich: Is `Event` needed here?

```suggestion
  onClickAway: (event: T) => void,
```

PR #867, opened 2020-01-08
feat(useBeforeUnload): allow passing a dirty function (#842)

### LINKED ISSUE #842: `useOnBeforeUnload()` on expensive dirty flag

I've considered the `useOnBeforeUnload` hook to warn if a doc has unsaved changes.
But it makes no sense because the dirty boolean flag is expensive to calculate (it's a deep comparison between serialized objects).

The way the hook currently works I have to recalculate the dirty flag much more often than necessary.

I'd like to propose that

1. the `useOnBeforeUnload` hook *always* registers a handler.
2. I can optionally provide a `dirty` *function* instead of a boolean
3. The handler, if called before window unload, calls `dirty()` if it's a function, but only at the actual time of unload where it's relevant.

That would reduce an expensive calculation to once when it's needed.
For simplicity, I think the approach can be the same for a boolean `dirty` flag since the handler won't take up noticeable resources.

I have considered flagging the `isDirty` on any doc change and avoiding re-calculation once it's `true`, but that doesn't make much sense because there's so many changes to the doc variable that are not relevant anymore after it's serialized (such as selection changes).

So it looks like I can either PR this for you or build my own hook.

bebbi:
# Description

Allow passing a function as dirty flag to `useBeforeUnload` hook.
Link documentation of an example how to use this with fast-changing dirty functions.

Closes #842 

## Type of change

# Checklist
- [x] (n.a.) Comment the code, particularly in hard-to-understand areas
- [x] Add documentation
- [x] (n.a.) Cover changes with tests

### DIFF
--- a/docs/useBeforeUnload.md
+++ b/docs/useBeforeUnload.md
@@ -5,12 +5,39 @@ React side-effect hook that shows browser alert when user try to reload or close
 
 ## Usage
 
+### Boolean check
+
 ```jsx
 import {useBeforeUnload} from 'react-use';
 
 const Demo = () => {
   const [dirty, toggleDirty] = useToggle(false);
   useBeforeUnload(dirty, 'You have unsaved changes, are you sure?');
+
+  return (
+    <div>
+      {dirty && <p>Try to reload or close tab</p>}
+      <button onClick={() => toggleDirty()}>{dirty ? 'Disable' : 'Enable'}</button>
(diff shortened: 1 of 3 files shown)


bebbi:
@streamich stumbled into another review, made an aesthetic fix and did a history-rewriting commit assuming you haven't pulled it yet.

PR #873, opened 2020-01-10
[WIP] Lock iOS Body on click

andrico1234:
# Description

This gets the absolute minimum working, but I want to get feedback before continuing. I'll also annotate the code for changes

## Type of change

# Checklist
- [ ] Add documentation

ezgif com-video-to-gif

### DIFF
--- a/src/useLockBodyScroll.ts
+++ b/src/useLockBodyScroll.ts
@@ -15,14 +15,32 @@ export function getClosestBody(el: Element | HTMLElement | HTMLIFrameElement | n
   return getClosestBody((el as HTMLElement).offsetParent!);
 }
 
+function preventDefault(rawEvent: TouchEvent): boolean {
+  const e = rawEvent || window.event;
+  // Do not prevent if the event has more than one touch (usually meaning this is a multi touch gesture like pinch to zoom).
+  if (e.touches.length > 1) return true;
+
+  if (e.preventDefault) e.preventDefault();
+
+  return false;
+}
+
 export interface BodyInfoItem {
   counter: number;
   initialOverflow: CSSStyleDeclaration['overflow'];
 }
 
+const isIosDevice =
+  typeof window !== 'undefined' &&
+  window.navigator &&
+  window.navigator.platform &&
+  /iP(ad|hone|od)/.test(window.navigator.platform);
+
 const bodies: Map<HTMLElement, BodyInfoItem> = new Map();
 
 const doc: Document | undefined = typeof document === 'object' ? document : undefined;
+
+let documentListenerAdded = false;
 
 export default !doc
   ? function useLockBodyMock(_locked: boolean = true, _elementRef?: RefObject<HTMLElement>) {}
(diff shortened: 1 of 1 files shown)


src/useLockBodyScroll.ts
```diff
@@ -15,15 +15,33 @@ export function getClosestBody(el: Element | HTMLElement | HTMLIFrameElement | n
   return getClosestBody((el as HTMLElement).offsetParent!);
 }
 
+function preventDefault(rawEvent: TouchEvent): boolean {
```
andrico1234: This function prevents the onTouchMove from performing any functionality. I could rename this for clarity, since the only time we use this functionality is on the onTouchMove event listener.

andrico1234: Something to also consider for this PR is how we handle use cases in which the onToucnMove has been changed on the client-side. This function will override this. 

Is this a cause for concern?

src/useLockBodyScroll.ts
```diff
@@ -15,15 +15,33 @@ export function getClosestBody(el: Element | HTMLElement | HTMLIFrameElement | n
...
+
+  return false;
+}
+
 export interface BodyInfoItem {
   counter: number;
   initialOverflow: CSSStyleDeclaration['overflow'];
 }
 
+const isIosDevice =
```
andrico1234: Let me know if there's already a helper to detect this

src/useLockBodyScroll.ts
```diff
@@ -15,15 +15,33 @@ export function getClosestBody(el: Element | HTMLElement | HTMLIFrameElement | n
...
+  typeof window !== 'undefined' &&
+  window.navigator &&
+  window.navigator.platform &&
+  /iP(ad|hone|od)/.test(window.navigator.platform);
+
 const bodies: Map<HTMLElement, BodyInfoItem> = new Map();
 
 const doc: Document | undefined = typeof document === 'object' ? document : undefined;
 
+let documentListenerAdded = false;
```
andrico1234: Not sure where we stand with mutable variables.

andrico1234: This variable may not even be necessary

src/useLockBodyScroll.ts
```diff
@@ -40,15 +58,33 @@ export default !doc
...
           if (bodyInfo) {
             if (bodyInfo.counter === 1) {
               bodies.delete(body);
-              body.style.overflow = bodyInfo.initialOverflow;
+              if (isIosDevice) {
+                body.ontouchstart = null;
+                body.ontouchmove = null;
+
+                if (documentListenerAdded) {
+                  document.removeEventListener('touchmove', preventDefault);
```
andrico1234: this removes the default touchmove event

src/useLockBodyScroll.ts
streamich: No, we don't have one.

andrico1234:
@streamich I've tested this in browserstack and it works, you can check the gif I added. Safari works fine too, although it worked previously. I'll go and check on a variety of other browsers and operating systems too, but do you have any other scenarios that you recommend I manually test?

src/useLockBodyScroll.ts
streamich: LGTM

andrico1234:
I'm on holiday for a couple of days, so when I'm back I'll add the unit tests.

andrico1234:
Incredible, thanks so much!

Does this mean I can be added to the list of contributors?

streamich:
> Does this mean I can be added to the list of contributors?

Yeah, it should happen automatically.

Issue #594, opened 2019-09-06
[Feature] useCookie


AlexandreAWE:
Hello guys,

What do you think about a useCookie hook ?

```js
const [foo, setFoo, deleteFoo] = useCookie('foo');

setFoo('value', 365); // setFoo(value:string, expirationDays:number) => void
deleteFoo()
```

Possible implentation : 
```typescript
export function useCookie(cookieName:string): [string, (value: string, days?: number) => void, () => void] {
  const [cookieValue, setCookieValue] = useState(() => {
    const v = document.cookie.match('(^|;) ?' + cookieName + '=([^;]*)(;|$)');
    return v ? v[2] : null;
  });

  const updateCookie = (value: string, days: number = 365) => {
    const d = new Date();
    d.setTime(d.getTime() + 24 * 60 * 60 * 1000 * days);
    document.cookie = cookieName + '=' + value + ';path=/;expires=' + d.toUTCString();
    setCookieValue(value);
  };

  const deleteCookie = () => {
    updateCookie('', -1);
    setCookieValue(null);
  };

  return [cookieValue, updateCookie, deleteCookie];
}
```

Inspired from : https://plainjs.com/javascript/utilities/set-cookie-get-cookie-and-delete-cookie-5/

Waiting for reviews.

Thanks

CloudPower97:
Hey @AlexandreAWE , I was about to open an issue myself with the same exact request.
I think this would be really useful.

@AlexandreAWE We might want to use [`js-cookie`](https://github.com/js-cookie/js-cookie) to handle this.

@streamich Let us know your thought about this :smile:

streamich:
Would be interesting also to listen and re-render on cookie changes, [see](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/cookies/onChanged).

streamich:
- [ ] return latest value of cookie
- [ ] function to set cookie
- [ ] function to delete cookie, possibly by ussing set function, and setting cookie to `null`
- [ ] listen for cookie changes and re-render
- [ ] some sensible mock of this hook for server-side rendering

ayush987goyal:
@AlexandreAWE @CloudPower97 @streamich Hey guys! Could you please review the above PR?

ayush987goyal:
@streamich  This issue can be closed. Merged with #878

PR #878, opened 2020-01-11
Add useCookie hook

ayush987goyal:
# Description

Issue #594 

## Type of change

# Checklist
- [x] Add documentation


ayush987goyal:
Suggested changes
- [x] return latest value of cookie
- [x] function to set cookie
- [x] function to delete cookie, possibly by ussing set function, and setting cookie to null
- [ ] listen for cookie changes and re-render (there is no event emitted for cookie change, the one which is present is only for extensions)
- [ ] some sensible mock of this hook for server-side rendering

docs/useCookie.md
```diff
@@ -0,0 +1,39 @@
...
+      <button onClick={deleteCookie}>Delete Cookie</button>
+    </div>
+  );
+};
+```
+
+## Reference
+
+```ts
+useCookie(cookieName: string);
```
streamich:
```suggestion
const [value, updateCookie, deleteCookie] = useCookie(cookieName: string);
```

src/useCookie.ts
```diff
@@ -0,0 +1,25 @@
+import { useState, useCallback } from 'react';
+import Cookies from 'js-cookie';
+
+const useCookie = (
+  cookieName: string
+): [string | null, (newValue: string, options?: Cookies.CookieAttributes) => void, () => void] => {
+  const [value, setValue] = useState<string | null>(() => Cookies.get(cookieName) || null);
```
streamich: Shall we use `undefined` instead of `null`, same as `js-cookie`?

ayush987goyal: I used `null` instead of `undefined` here because we are explicitly setting the value to a nullish value in `deleteCookie`. It is very rare to programmatically assign any value to undefined.

streamich: Can you set it to `undefined` in `deleteCookie`?

ayush987goyal: Yes I can, but I am not sure if it would be the correct approach. If we have followed that convention all across, I can go ahead and make it `undefined`. Do you want me to go ahead and do that?

streamich: I was curious if there is a reason it has to be `setValue(null)` in `deleteCookie`, because you were saying: "...because we are explicitly setting the value to a nullish value in `deleteCookie`". Its up to you how to leave it.

ayush987goyal: We do have to set it to `null` so as to cause a rerender since we do not have a listener for cookie change. I think it is better if we use `null` instead of `undefined` here.

Issue #834, opened 2019-12-13
useLocalStorage should support JSON.stringify replacer's and reviver's.


cosullivan:
I am using Typescript and have a Type that contains Set<T>'s. When using useLocalStorage this fails because Set<T>'s can't be serialized with JSON.stringify and as such they disappear from my type.

The solution to having JSON.stringify work on Set's is to use the replacer function on JSON.stringify and the reviver function on JSON.parse. 

It would be good if useLocalStorage could take in these functions as optional parameters to be used during stringify and parse or alternatively callbacks that control the serialization and deserialization process.

streamich:
Currently, `JSON.stringify` is an internal implementation, exposing replacer and reviver would expose that internal detail. What are the benefits/drawbacks of normalizing or serializing the data before passing it to the `useLocalStorage` hook?

cosullivan:
That's fair enough about not wanting to expose the internal implementation, the issue is that the internal implementation can silently fail.

I do currently have a custom hook which wraps the useLocalStorage hook and performs the serialization before and after and that works fine, but it's not as succinct as just being able to pass in a serializer & deserializer function. 

Given that the constraint/api on localStorage is to use string values perhaps having optional parameters one for the serializer which is _T => string_ and once for the deserializer which is _string => T_ and the default impementation of those is JSON.stringify and JSON.parse?

streamich:
Yeah, I like the idea of serializer to be an optional argument.

ayush987goyal:
@streamich @cosullivan Hey guys! Can you please review the above PR?

cosullivan:
This API change will work nicely for me.

Thanks, Cain.

ayush987goyal:
@streamich This can be closed. Merged with #896

PR #896, opened 2020-01-15
feat: add serializer/deserializer option to useLocalStorage

ayush987goyal:
# Description

1. #834 
2. Add tests for `useLocalStorage` hook

## Type of change

# Checklist
- [x] Add documentation

### DIFF
--- a/src/useLocalStorage.ts
+++ b/src/useLocalStorage.ts
@@ -27,7 +27,7 @@ const useLocalStorage = <T>(
   const [state, setState] = useState<T>(() => {
     try {
       const localStorageValue = localStorage.getItem(key);
-      if (localStorageValue) {
+      if (localStorageValue !== null) {
         return deserializer(localStorageValue);
       } else {
         initialValue && localStorage.setItem(key, serializer(initialValue));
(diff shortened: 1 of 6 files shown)


src/useLocalStorage.ts
```diff
@@ -1,22 +1,37 @@
...
+  const serializer = options ? (options.raw ? String : options.serializer) : JSON.stringify;
+  const deserializer = options ? (options.raw ? String : options.deserializer) : JSON.parse;
+
   const [state, setState] = useState<T>(() => {
     try {
       const localStorageValue = localStorage.getItem(key);
-      if (typeof localStorageValue !== 'string') {
-        localStorage.setItem(key, raw ? String(initialValue) : JSON.stringify(initialValue));
-        return initialValue;
+      if (localStorageValue) {
```
streamich: This probably should check for `undefined`, in case `localStorageValue` is an empty string.

```suggestion
      if (localStorageValue !== undefined) {
```

For example

```js
useLocalStorage('foo', '');
```

ayush987goyal: Woops just missed that. On that note, it should check for `null` instead of `undefined` since `localStorage.getItem()` returns with a type of `string | null`

Issue #597, opened 2019-09-09
withHooks HOC to decouple components from hooks


malerba118:
## Is your feature request related to a problem? Please describe.

With the introduction of the Hooks API in React 16.8, developers have been shifting away from Higher Order Components and Render Props patterns to handle cross-cutting concerns amongst React components. Prior to Hooks, [recompose](https://github.com/acdlite/recompose) offered a robust library of HOC's with similar functionalities to the Hooks API.
Since then, the creator of recompose and the React team have been advising developers make the switch from recompose to Hooks as they feel Hooks solve all of the issues recompose aimed to solve, but avoid issues inherent to recompose such as excessive tree nesting.

However, Hooks are not perfect either. And there's been a lot of discussion around one issue in particular. The recommended way to implement Hooks is to couple them with the components that depend on them. By doing this, we complicate things such as render bail-outs and testing because now components bake-in hook logic and explicitly depend on these hooks as their data/logic providers. One of the nice things about recompose HOC's is that they inject similar functionality through _props_, meaning the components are not explicitly dependent on their data providers. This [issue thread](https://github.com/acdlite/recompose/issues/756#issuecomment-435576893) on the recompose repo discusses much of the motivation for this library.

## Describe the solution you'd like
```jsx
import React, { useState } from "react";
import ReactDOM from "react-dom";

const pick = (obj, fields) => {
  let picked = {};
  if (obj instanceof Object && fields instanceof Array) {
    fields.forEach(field => {
      picked[field] = obj[field];
    });
  }
  return picked;
};

const withHooks = (mapHooksToProps, { frozenProps = [] } = {}) => {
  return WrappedComponent => {
    return props => {
      let overrideablePropsObj = mapHooksToProps(props);
      let frozenPropsObj = pick(overrideablePropsObj, frozenProps);
      return (
        <WrappedComponent
          {...overrideablePropsObj}
          {...props}
          {...frozenPropsObj}
        />
      );
    };
  };
};

const Counter = props => {
  return (
    <div>
      <div>Counter: {props.count}</div>
      <button onClick={props.increment}>Increment</button>
      <button onClick={props.decrement}>Decrement</button>
    </div>
  );
};

const mapHooksToProps = props => {
  let [count, setCount] = useState(0);

  return {
    count,
    increment: () => setCount(prev => prev + 1),
    decrement: () => setCount(prev => prev - 1)
  };
};

const EnhancedCounter = withHooks(mapHooksToProps)(Counter);

[...]

PR #903, opened 2020-01-16
feat: add withHook higher order component

ayush987goyal:
fix: path for useError docs

# Description

1. #597 
2. Fix for `useError` docs path.

* I am not confident about the `types` of this `withHook` HOC. Would need help in getting those correct.
* Unit tests are not added since they would require some thinking.

## Type of change

# Checklist
- [x] Add documentation


src/withHook.tsx
```diff
@@ -0,0 +1,13 @@
...
+
+function withHook<T>(useHook: passedHook<T>) {
+  return (WrappedComponent: React.ComponentType<T>) => (props: React.Props<T>) => {
+    const propsToMapFromHook = useHook(props);
+
+    return <WrappedComponent {...propsToMapFromHook} {...props} />;
+  };
+}
+
+export default withHook;
```
ayush987goyal: Just noticed, `React.Props` is deprecated. Maybe we can just say `props: T` instead?

src/withHook.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+
+type passedHook<T> = (props: React.Props<T>) => any;
```
malerba118: Because we're destructuring the return value and passing it as props this should probably restrict them to returning an object or nothing `type passedHook<T> = (props: React.Props<T>) => object | undefined;`

malerba118:
I'd recommend picking apart react-redux types for `connect` because it should be almost identical to what we want to do where `passedHook` is equivalent to `mapStateToProps` https://github.com/DefinitelyTyped/DefinitelyTyped/blob/5344bfc80508c53a23dae37b860fb0c905ff7b24/types/react-redux/index.d.ts

src/withHook.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+
+type passedHook<T> = (props: React.Props<T>) => object | undefined;
```
FlickerLogicalStack: with such implementation you can't pass hook like `const useNumber = ({ number }: { number: number }) => number;` because `number` not in `object` set

ayush987goyal:
@streamich I think this basic implementation is currently in usable state and can be merged. We can do incremental changes to it if necessary.

malerba118:
My only other thought is should we put a forwardRef around this so that refs will point to the passed component

ayush987goyal:
That's a good catch. I am not sure if typical HOCs wrap themselves with `forwardRef`, but we can do it if we want to support that.

ayush987goyal:
@malerba118 How does this implementation look?

```ts
import React, { forwardRef } from 'react';

type passedHook<T> = (props: React.Props<T>) => object | undefined;

function withHook<T>(useHook: passedHook<T>) {
  return (WrappedComponent: React.ComponentType<T>) =>
    forwardRef((props: React.PropsWithChildren<T>, ref) => {
      const propsToMapFromHook = useHook(props);

      return <WrappedComponent ref={ref} {...propsToMapFromHook} {...props} />;
    });
}

export default withHook;

```

malerba118:
Looks good! Thanks!

FlickerLogicalStack:
I think where is 2 problems:
1.  `passedHook` return type constaint must be anything
2. We need mapper like `mapHookToProps` (with default `(hookReturnData) => returnData`) for independence of return value of hook and props shape of WrappedComponent

malerba118:
I do prefer the name mapHookToProps to useHook because it’s more familiar terminology for people. But I don’t think it should have a default value because it’s a required argument. Also I don’t think it should have a return type of any because say mapHookToProps returned a number, then withHook wouldn’t know what prop name to pass that number to. An object is necessary because the keys indicate the prop names and the values indicate the prop values.

ayush987goyal:
Maybe we can make the `mapHookToProps` an optional argument? This way if your hook does indeed returns any primitive value, it can be mapped to a prop through `mapHookToProps`. And if you don't pass that, it will behave like the current implementation. What say?

FlickerLogicalStack:
@malerba118

> Also I don’t think it should have a return type of any because say mapHookToProps returned a number, then withHook wouldn’t know what prop name to pass that number to.

The `passedHook` itself should not have any constraints since it is an independent code. There should be constraints to the mapper return value. Because the main role of mapper to receives the hook value at the input and adapts it to the shape of the component's props.

ayush987goyal:
@FlickerLogicalStack @malerba118 How does this look?
```ts
import React, { forwardRef } from 'react';

type passedHook<T> = (props: React.Props<T>) => any;
type HookToPropsMapper = (hookValues: any) => object;

function withHook<T>(useHook: passedHook<T>, mapHookToProps: HookToPropsMapper = hookValues => hookValues) {
  return (WrappedComponent: React.ComponentType<T>) =>
    forwardRef((props: React.PropsWithChildren<T>, ref) => {
      const hookValues = useHook(props);
      const propsToMapFromHook = mapHookToProps(hookValues);

      return <WrappedComponent ref={ref} {...propsToMapFromHook} {...props} />;
    });
}

export default withHook;

```

FlickerLogicalStack:
@ayush987goyal
This code will work in runtime (I hope so 😀), but it is absolutely type unsafe.

I can suggest _nottrue_ implementation with **strange** API, but it can be static validated

```tsx
type Hook<TProps, TReturnType> = (props: TProps) => TReturnType;

function withHook<THookProps, THookReturnType>(
  hook: Hook<THookProps, THookReturnType>
) {
  return <TComponentProps, TMapperReturnType extends TComponentProps>(
    Component: React.ComponentType<TComponentProps>,
    mapHookToProps: (
      hook: THookReturnType,
      props: TComponentProps
    ) => TMapperReturnType
  ) => {
    return (
      props: THookProps & TComponentProps
    ) => {
      const hookValue = hook(props);

      return <Component {...mapHookToProps(hookValue, props)} />;
    };
  };
}
```

usage:
```tsx
const useNumber = ({ number }: { number: number }) => number;

const Component = (_: { b: string; c: null }) => null;

const Wrapped = withHook(useNumber)(Component, (hook, props) => ({
  b: hook.toString(),
  c: props.c,
}));
```

Now `Wrapped` component will require props `a`& `b` & `c` and `mapHookToProps` will require return equal shape of `Component` props

(I totally realise that is very **so so** suggestion )

FlickerLogicalStack:
The main problem that TS can't (or just i can't) infer generics from nested functions calls __from the bottom up__.
Therefore, I we can't do normal API and static validation  😞

ayush987goyal:
@FlickerLogicalStack That's pretty cool implementation. Adding a revision.

FlickerLogicalStack:
@ayush987goyal

There one more thing to that implementation:

```tsx
interface PlainObjectShape extends {};

type Hook<TProps extends PlainObjectShape, TReturnType> = (props: TProps) => TReturnType;
```

`Hook`'s TProps constraint must be plain object.
(Simple `extends {}` or `extends {[s: string]: any}` will not work)

P.S.
Also I thing type naming `Hook` is better because we need `PascalCase` name (as usually) and `passed` is unnecessary trait.
Anyway these are trifle 😁

FlickerLogicalStack:
@streamich, @malerba118 do you agree with this implementation?

ayush987goyal:
Why do we need to add a `PlanObjectShape` constraint? The `TProps` type is nothing but the actual props. Now that I think about it, `THookProps` should just be TComponentProps` since these will be passed from the parent component using this wrapped component.

FlickerLogicalStack:
@ayush987goyal because you pass `THookProps ` through JSX, there are no options through which structure to pass data

FlickerLogicalStack:
@ayush987goyal 
`PlanObjectShape` is just constraint of `{}`, because if `props: THookProps & TComponentProps` something like `(number & { b: string; c: null })` === `never`

Example ([Playground](http://www.typescriptlang.org/play/?jsx=2&ssl=35&ssc=26&pln=35&pc=39#code/JYWwDg9gTgLgBAJQKYEMDG8BmUIjgcilQ3wG4Aoc4AOxiSk3STgAUAbFGgeQCMArJBgDKACxRhmAb3Jw4AbQDOALjgKYUGgHMAuipTUAnhVmKV1AK4ge9XXGpIAbvQoBfCuRgGJcABIQIANYAPAAqLDhgCgA0cCHIMOZQ1CFeSAB8cAC8cAAUYBHKseEQkQCUWRlxSAlJKRLumObUGMAQ1HAA7sAwIn6BoX0BxZFwSAAedNQAJgqsHNz8gjCi4kgxIYPxicmpaTkycCL+ASqDA8fD0bGb1dt16eTl0rJENe2hAMK4kPa0l+sAWXEEigW1qqVGEyQ01mIS+4Da0Jglz2B1k8J+SJUyHQMAAdBjEbR7p9vkTkQU0lE0XAQOJBiEIJcVPtZGzDscVBtjmCdhJqezZPkSoU4WTfhSRTTyplKkCwCDefcDjKMs82a9trkaUKClzBpc4AAyWKEiWXaUVODq9loNpqDmBABqKDY5mY2SOgTyBVK7kFcE1STgQTNSOteMjdLADKZBRyXoCLrdazgwrKLjgAHo0sY2W4DgWXJQ7dQHeYFEgAEIoKYAOUs1igWVyAH0zI36Kq4ABGCil8uVgDi-nrneb2Ry7etFis9A7c6gLm7fZL9vgYdoLanKkkcB4KjUGmomlIcDQHbYbDgy6tFiv7gH8AA6lBgUgpi2uj1BjkK0gRwgMdF1KHJNxgGIE2OGJ0wUbscnVA9HQCPEYAgIR1C0HJSgFc8VFgvE0GpZc-UoIJX3fKYDh4TJJHwHt8GLWQ0Fo+82CYuxx1ontixzIA)):
```tsx
const useBadNumber = (_: number) => 1;
const useGoodNumber = (_: {number: number}) => 1;

const Component = (_: { b: string; c: null }) => null;

const Wrapped = withHook(useBadNumber)(Component, (hook, props) => ({
  b: hook.toString(),
  c: props.c,
}));
```

without constaint:
image
with constaint:
image


Just better TS API.

P.S.
`PlainObjectShape` must be
```tsx
interface PlainObjectShape {
  [s: string]: any;
  [s: number]: never;
};
```

malerba118:
I'm not a fan of this. There's no way to pass arguments to the hook.
```jsx
const Wrapped = withHook(useState)(Component, (hook, props) => ({
  
}));
```
There's no way to pass a default state value to this ^^^. It just receives props as the default state.

I think a better way is this

```jsx
const Wrapped = withHook((props) => {
  const hook =  useNumber(10, props.options)
  return {
    b: hook.toString(),
    c: props.c,
  }
})(Component);
```
I think this is better because it lets you pass arguments to the custom hook and it's more concise

streamich:
I'm thinking about an implementation where one can easily provide all arguments to the hook. In the current implementation as I understand the hook receives a single argument, but would be nice to any signature of arguments, for example, something like this:

```ts
const Wrapped = withHook<typeof useNumber>(Component, (props) => {
  const [value, controls] = useNumber(1, 0, 10);
  return {
    value,
    controls,
  };
});
```

What do you think?

ayush987goyal:
That's a clean implementation. I had a similar line of thought at the start. The second argument to `withHook` should be another `hook` that will receive `props`. So the above example can also be written as:
```ts
function useSomeHook(props) {
	const [value, controls] = useNumber(1, 0, 10);
	return {
	    value,
	    controls,
	}; // Extra props that will be merged into the props of Component.
}

const Wrapped = withHook(Component, useSomeHook);
```

Shall we go ahead with this this?

streamich:
> Shall we go ahead with this this?

Yes, this looks good to me.

ayush987goyal:
@streamich  As discussed in previous comments, do we include `mapHookToProps` in our implementation?
As far as I am concerned, we might not need it since with this implementation, you can always write a custom hook that will use your hook and map the props as required. What do you think?

FlickerLogicalStack:
@ayush987goyal, so, what about this case (based on the current implementation):
```TS
const useArray = () => [11, 22, 33];

const Component = (props: { number: number }) => null;

const Wrapped = withHook(Component, useArray);

<Wrapped number={1} />
```

As far as I understand calling of `Component` will be with `{ number: 1, 0: 11, 1: 22, 2: 33 }` props.
Maybe we need another constraint instead of `object`?

ayush987goyal:
@FlickerLogicalStack I agree to the above analogy. I could hardly think of a good way to implement the types for this. It would be helpful if you had any specific suggestion that we could discuss.

FlickerLogicalStack:
@ayush987goyal, the main problem in fact, that you can't safety pass any hook output to props of any component.
Even if we replace `object` in `type passedHook<T> = (props: T) => object;` with another constraint, it will be just a deception in terms of runtime logic (we can still get props like `{ number: 1, 0: 11, 1: 22, 2: 33 }`).
Runtime checks on shape (_we pass an object or an array?_) equality will also so creepy.
The only type-safe and runtime-safe suggestion that i can propose is my _old_ idea with `mapHookToProps`.
as they say "back to the future."😇

PR #908, opened 2020-01-17
feat: `useMap`: allow resetting with provided value other then initial

mattleonowicz:
# Description

`useMap`: Add `setAll` to allow providing entirely new object

## Type of change

# Checklist
- [x] Add documentation

### DIFF
--- a/src/useMap.ts
+++ b/src/useMap.ts
@@ -2,6 +2,7 @@ import { useState, useMemo, useCallback } from 'react';
 
 export interface StableActions<T extends object> {
   set: <K extends keyof T>(key: K, value: T[K]) => void;
+  setAll: (newMap: T) => void;
   remove: <K extends keyof T>(key: K) => void;
   reset: () => void;
 }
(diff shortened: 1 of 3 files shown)


mattleonowicz:
Potentially if someone uses `reset` directly in an event handler like `onClick={reset}` then this would break his code, as the event would be passed as new object here. I'm not sure if that qualifies as a breaking change. Other solution would be to add a separate method like `resetWith`. Let me know what you think, I can change it accordingly.

src/useMap.ts
```diff
@@ -27,7 +27,7 @@ const useMap = <T extends object = any>(initialMap: T = {} as T): [T, Actions<T>
           return rest as T;
         });
       },
-      reset: () => set(initialMap),
+      reset: (newMap = initialMap) => set(newMap),
```
streamich: In this case, should you also keep track of the "new" `initialMap`?

```js
reset: (newMap = initialMapRef.current) => {
  initialMapRef.current = newMap;
  set(newMap);
},
```

streamich:
Maybe instead of overloading `reset` function, we can add `put` function?

mattleonowicz:
Yeah, maybe it would be better.
 In other languages it's common to see `put` and `putAll` for a Map,  so maybe we should use `setAll`, since we already have `set`.

Issue #206, opened 2019-04-03
useMethods


atomiks:
Similar to: https://github.com/pelotom/use-methods

Neat alternative to `useReducer`

## Hook

```tsx
function useMethods(createMethods, initialState) {
  const reducer = useMemo(
    () => (state, action) => {
      return createMethods(state)[action.type](...action.payload);
    },
    [createMethods]
  );
  const [state, dispatch] = useReducer(reducer, initialState);
  const actionTypes = Object.keys(createMethods(initialState));
  const wrappedMethods = useMemo(() => {
    return actionTypes.reduce((acc, type) => {
      acc[type] = (...payload) => dispatch({ type, payload });
      return acc;
    }, {});
  }, actionTypes);

  return [state, wrappedMethods];
}
```

## Usage

```tsx
const initialState = {
  count: 0,
};

function createMethods(state) {
  return {
    reset() {
      return initialState;
    },
    increment() {
      return { ...state, count: state.count + 1 };
    },
    decrement() {
      return { ...state, count: state.count - 1 };
    },
  };
}

function Counter() {
  const [state, methods] = useMethods(createMethods, initialState);

  return (
    <>
      Count: {state.count}
      <button onClick={methods.increment}>+</button>
      <button onClick={methods.decrement}>-</button>
    </>
  );
}
```

wardoost:
This looks pretty neat indeed. Is there also a way to use async methods with this?

atomiks:
> Is there also a way to use async methods with this?

Perhaps return a function that receives the methods object like this:

```jsx
const createMethods = state => ({
  incrementAsync: () => async methods => {
    await new Promise(done => setTimeout(done, 1000));
    methods.increment();
  }),
  increment: () => ({ ...state, count: state.count + 1 })
});
```

So you will call the sync reducers from the async methods.

```jsx

function useMethods(createMethods, initialState) {
  const reducer = useMemo(
    () => (state, action) => {
      return createMethods(state)[action.type](...action.payload);
    },
    [createMethods]
  );
  const [state, dispatch] = useReducer(reducer, initialState);
  const methods = createMethods(initialState);
  const actionTypes = Object.keys(methods);
  const wrappedMethods = useMemo(() => {
    return actionTypes.reduce((acc, type) => {
      acc[type] =
        typeof methods[type]() === 'function'
          ? (...payload) => methods[type](...payload)(acc)
          : (...payload) => dispatch({ type, payload });
      return acc;
    }, {});
[...]

PR #910, opened 2020-01-18
feat: add useMethods state hook

ayush987goyal:
# Description

1. #206 

## Type of change

# Checklist
- [x] Add documentation

### DIFF
--- a/docs/useMethods.md
+++ b/docs/useMethods.md
@@ -1,0 +1,48 @@
+# `useMethods`
+
+React hook that simplifies the `useReducer` implementation.
+
+## Usage
+
+```jsx
+import { useMethods } from 'react-use';
+
+const initialState = {
+  count: 0,
+};
+
+function createMethods(state) {
+  return {
+    reset() {
+      return initialState;
+    },
+    increment() {
+      return { ...state, count: state.count + 1 };
+    },
+    decrement() {
+      return { ...state, count: state.count - 1 };
+    },
+  };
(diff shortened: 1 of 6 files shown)


docs/useMethods.md
```diff
@@ -0,0 +1,48 @@
...
+      return initialState;
+    },
+    increment() {
+      return { ...state, count: state.count + 1 };
+    },
+    decrement() {
+      return { ...state, count: state.count - 1 };
+    },
+  };
+}
```
streamich: What do you think about providing `state` to every method?

```ts
const methods = {
  reset: state => () => initialState,
  increment: state => () => { ...state, count: state.count + 1 },
  decrement: state => () => { ...state, count: state.count - 1 },
};
```

ayush987goyal: That is an interesting idea but I feel that the current design looks better since the methods will have closure over `state` anyway. Otherwise, the definition of `methods` might be tricky to get heads around. Right now, it is just a function that returns some methods. Thoughts?

ayush987goyal:
This also has `eslint` fixes for `useMethods` for #947

docs/useMethods.md
streamich: I'm fine with either approach. Just wanted to hear your thoughts, because the current implementation on every state change has to generate a new object `{ reset, increment, ... }`, whereas in the other implementation only one function is generated.

PR #942, opened 2020-02-01
Make useLocalStorage set local storage on unmount

roninjin10:
# Description

useLocalStorage updates local storage within the react lifecycle with a useEffect.  This works well since
the update will happen immediately and the more expensive task of writing to local storage is done after the component already updates to reflect the change.

The problem is when a component unmounts,local storage does not get written to.  One common example of this happening is if we try to write to local storage while also going to a different page in our app.

This change makes the hook aware of whether it has unmounted or not and if it has unmounts, it will go ahead and write directly to local storage.

## Type of change

# Checklist
- [ ] Add documentation

### DIFF
--- a/src/useLocalStorage.ts
+++ b/src/useLocalStorage.ts
@@ -49,7 +49,7 @@ const useLocalStorage = <T>(
     };
   });
 
-  const setItem = (newState: T) => {
+  const setLocalStorage = (newState: T) => {
     try {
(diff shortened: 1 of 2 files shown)


roninjin10:
Making a new more robust pull request to address some other open issues with useLocalStorage

streamich:
@roninjin10 please make your PR into `v14.0` branch.

PR #958, opened 2020-02-10
pr/useWindowSize callback

### LINKED ISSUE #915: useWindowSize callback
Add ability to trigger a callback when the size is updated.

```
const { ... } = useWindowSize(() => {
...callback on resize...
});
```

morgsmccauley:
# Description

Resolves #915 

Still trying to wrap my head round `requestAnimationFrame` so let me know if this PR is fine. Can update docs if all is well :)

## Type of change

# Checklist
- [x] Add documentation

### DIFF
--- a/src/useWindowSize.ts
+++ b/src/useWindowSize.ts
@@ -3,11 +3,16 @@ import { useEffect } from 'react';
 import useRafState from './useRafState';
 import { isClient } from './util';
 
-const useWindowSize = (initialWidth = Infinity, initialHeight = Infinity) => {
+const useWindowSize = (initialWidth = Infinity, initialHeight = Infinity, callback?: CallableFunction) => {
   const [state, setState] = useRafState<{ width: number; height: number }>({
     width: isClient ? window.innerWidth : initialWidth,
     height: isClient ? window.innerHeight : initialHeight,
   });
+
+  useEffect(() => {
+    if (!callback) return;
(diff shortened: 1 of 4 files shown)


src/useWindowSize.ts
```diff
@@ -3,11 +3,14 @@ import { useEffect } from 'react';
...
-    height: isClient ? window.innerHeight : initialHeight,
-  });
+const useWindowSize = (initialWidth = Infinity, initialHeight = Infinity, callback?: CallableFunction) => {
+  const [state, setState] = useRafState<{ width: number; height: number }>(
+    {
+      width: isClient ? window.innerWidth : initialWidth,
+      height: isClient ? window.innerHeight : initialHeight,
+    },
+    callback
+  );
```
streamich: Instead of adding `callback` to `useRafState` can we add here on the next line a `useEffect` block like this:

```ts
useEffect(() => {
  if (!callback) return;
  callback(state);
}, [state.width, state.height]);
```

streamich:
I've switched base branch of this PR to `v14.0` as this might be a substantial change in behaviour for some users.

src/useWindowSize.ts
morgsmccauley: Good call, updated 😄

morgsmccauley:
Happy to re-open if you still think there is need 😄

PR #959, opened 2020-02-11
Add useBobo hook

bignimbus:
# Description

DJ Bobo (aka DJ BoBo) took the world, but mostly Europe, by storm in 1994 with ["There is a Party."](https://en.wikipedia.org/wiki/There_Is_a_Party_(DJ_BoBo_song))  The React community has largely ignored this phenomenon until now.  The `useBobo` hook exposes a simple interface through which React developers can rapidly expose their users to OFFICIAL audio and video versions of DJ Bobo's magnum opus.  Changes are documented and tested thoroughly.

`CONTRIBUTING.md` was also updated with the correct paths to the test and story directories.  Also, vim users will note that their swap files are added to `.gitignore`.

## Type of change

# Checklist
- [x] Add documentation
- [x] Beats are fire

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -77,3 +77,10 @@ src/parser.ts
 
 storybook-static/
 package-lock.json
+
+# vim swap files
+[._]*.s[a-v][a-z]
+[._]*.sw[a-p]
+[._]s[a-rt-v][a-z]
+[._]ss[a-gi-z]
+[._]sw[a-p]
--- a/src/useBobo.ts
+++ b/src/useBobo.ts
@@ -1,0 +1,18 @@
+const SPOTIFY_URL = 'https://open.spotify.com/track/1jGATZ2aFc6yL5ycanq0iO';
+const YOU_TUBE_URL = 'https://www.youtube.com/watch?v=nn5hCEMyE-E';
+
+interface BoboAttrs {
+  spotifyUrl: string;
+  youTubeUrl: string;
+}
+
+const BOBO_ATTRS: BoboAttrs = {
+  spotifyUrl: SPOTIFY_URL,
+  youTubeUrl: YOU_TUBE_URL,
+};
+
+export function useBobo(): BoboAttrs {
+  return BOBO_ATTRS;
+}
+
+export default useBobo;
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -31,8 +31,8 @@ This library is a collection of React hooks so a proposal for a new hook will ne
 
 ### Creating a new hook
 
-1. Create `src/useYourHookName.ts` and `src/__stories__/useYourHookName.story.tsx`, run `yarn start` to start the storybook development server and start coding your hook
-1. Create `src/__tests__/useYourHookName.test.ts`, run `yarn test:watch` to start the test runner in watch mode and start writing tests for your hook
+1. Create `src/useYourHookName.ts` and `stories/useYourHookName.story.tsx`, run `yarn start` to start the storybook development server and start coding your hook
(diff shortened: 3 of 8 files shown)


src/useBobo.ts
```diff
@@ -0,0 +1,18 @@
...
+  youTubeUrl: string;
+}
+
+const BOBO_ATTRS: BoboAttrs = {
+  spotifyUrl: SPOTIFY_URL,
+  youTubeUrl: YOU_TUBE_URL,
+};
+
+export function useBobo(): BoboAttrs {
+  return BOBO_ATTRS;
```
streamich: It all looks good, but this PR states "provides the up-to-date links", but these links are hard-coded, not up-to-date. Another way to look at it is from our `CONTRIBUTING.md`:

> .. a proposal for a new hook will need to utilize the React Hooks API internally to be taken into consideration.

Stated another way, a React hook internally needs to use at least another React hook. Otherwise it is just a JavaScript function, not a React hook. If this hook would use other hooks to somehow fetch the up-to-date URLs, instead of them being hard-coded in the code, then this would be a real hook and a really good one, exactly something this library always needed and was missing all this time.

.gitignore
```diff
@@ -77,3 +77,10 @@ src/parser.ts
 
 storybook-static/
 package-lock.json
+
+# vim swap files
+[._]*.s[a-v][a-z]
+[._]*.sw[a-p]
+[._]s[a-rt-v][a-z]
+[._]ss[a-gi-z]
+[._]sw[a-p]
```
streamich: This looks like key strokes of somebody trying to exit vim.

src/useBobo.ts
bignimbus: > exactly something this library always needed and was missing all this time.

So glad you share the vision.  I appreciate the suggestions!  I originally intended for the links to be updated via the normal release cycle, e.g. handling URL rot via a patch release or something.  I worry that hardcoding a reference to a hosted service (BaaS, Bobo as a service, if you will) is more brittle as part of a utility library.  Perhaps we should consider making this hook consume `useState`, which would then conditionally return an audio or video link?

```jsx
// rough draft
const [{ medium, url }, updateBoboMedium] = useBobo('audio');
return (
  <div>
    <code>{ url }</code>
    <button onClick={() => updateBoboMedium('video')}>
      Render { medium === 'audio' ? 'Video' : 'Audio' } URL
    <button>
  </div>
);
```

bignimbus: Third option: I could create an npm module with the sole purpose of storing the latest URLs to official "There is a Party" streams in a JSON file.  That module could be declared as a dependency.  Pros: updates would be patch releases, so we wouldn't need to explicitly update `react-use` in the event that a URL needs to be mutated.  Cons: module tax, plus it seems like a lot of work for a feature that will only need to be updated once every few years, if ever.

streamich: The proposed `useState` alternative seems forced. We need something better, for example, it could use the `useAudio` hook:

```ts
const useBobo = () => useAudio({src});
```

bignimbus: @streamich that would be the ultimate realization of this idea.  My only concern with creating an implementation of `useAudio` would be copyright.  It's not typical for a commercial artist to host DRM-free versions of their recordings that can be distributed by any random software developer.  We could try to get permission from DJ Bobo himself to grant a license to distribute "There is a Party" under `react-use`'s license if and only if it's done through the `useBobo` API.  DJ Bobo's art has given us so much, but I will ask him for this favor for the good of the project.  Without permission, though, I don't think we can ship something like this.

FlickerLogicalStack:
Sorry, i probably can't understand the joke, but why we need hook that provide URLs to the noname singer.
And will it be possible after this PR to do others with suggestions of links to pewdiepie, rammstein, ets? 😅

PR #968, opened 2020-02-13
feat(useSet): add toggle a method

dispix:
# Description

This adds a `toggle` method to the `useSet` hook. It creates a handy wrapper on top of `add` and `remove`.

## Type of change

# Checklist
- [x] Add documentation

### DIFF
--- a/src/useSet.ts
+++ b/src/useSet.ts
@@ -3,6 +3,7 @@ import { useState, useMemo, useCallback } from 'react';
 export interface StableActions<K> {
   add: (key: K) => void;
   remove: (key: K) => void;
+  toggle: (key: K) => void;
   reset: () => void;
 }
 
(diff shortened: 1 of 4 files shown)


stories/useSet.story.tsx
```diff
@@ -13,6 +13,7 @@ const Demo = () => {
       <button onClick={() => remove('hello')} disabled={!has('hello')}>
         Remove 'hello'
       </button>
+      <button onClick={() => toggle('hello')}>{has('hello') ? 'Remove' : 'Add'} 'hello'</button>
```
wardoost: While it's a neat way to demonstrate the `has` method in code it wasn't immediately clear this button was the toggle button from looking at the UI. In my opinion the button label `Toggle 'hello'` is more clear. The `has` method could be demonstrated in a different way using an input field for example.

src/useSet.ts
```diff
@@ -13,14 +14,13 @@ export interface Actions<K> extends StableActions<K> {
...
-      add: item => setSet(prevSet => new Set([...Array.from(prevSet), item])),
-      remove: item => setSet(prevSet => new Set(Array.from(prevSet).filter(i => i !== item))),
-      reset: () => setSet(initialSet),
-    }),
-    [setSet]
-  );
+  const stableActions = useMemo<StableActions<K>>(() => {
+    const add = (item: K) => setSet(prevSet => new Set([...Array.from(prevSet), item]));
+    const remove = (item: K) => setSet(prevSet => new Set(Array.from(prevSet).filter(i => i !== item)));
+    const toggle = (item: K) => (set.has(item) ? remove : add)(item);
```
wardoost: The toggle does not add `hello` back to the set in the story for me. I think this has to do with the `set` used in this function refers to a previous value. Using the following toggle function results in the desired toggle behaviour

```ts
const toggle = (item: K) =>
  setSet(prevSet => {
    if (prevSet.has(item)) {
      return new Set(Array.from(prevSet).filter(i => i !== item));
    } else {
      return new Set([...Array.from(prevSet), item]);
    }
  });
```

The test passes for both versions so I'm not sure if the test is testing the right things?

dispix: Indeed! That seemed so straightforward that I didn't test it manually enough, in the end closures are always trickier than expected 😄 

Fixed it and updated the test to catch this kind of behaviour.

stories/useSet.story.tsx
dispix: I simplified the example so it's focused on the use of `toggle` 🙂

wardoost:
Thanks @dispix 🙌

PR #972, opened 2020-02-15
use fast-deep-equal for deep comparisons

streamich:


sKopheK:
fast-deep-equal causes "Maximum call stack size exceeded" error when object has circular dependencies

PR #978, opened 2020-02-16
update the sessionStorage to follow the same signature and behavior of localStorage

iamrommel:
# Description

sessionStorage and localStorage should have similar behavior, their only difference is where the data is being stored and it's persistence which should be abstracted for this library.

## Type of change
- [ ] **Breaking change**, The signature/arguments follows the useLocalStorage, this is breaking changes, because the previous last argument is just a boolean "raw", which i change it to object "options"
- [ ] New feature, Added the sessionStorage unit test
- [ ] New feature, Update the docs that describe about the config changes, for raw, serializer, de-serializer
- [ ] Bug fix, Update the useEffect implementation to use the state as dependency (previously there is no dependency at all)
- [ ] Bug fix, Removed the single-quote in lint script, to properly find the correct files and folder, this is not related to the changes, but the lint command does not find the files with single qoute, i tried with bash and command terminals both does not work there (we can discuss more on this if there is a need to)

 
# Checklist
- [x] Add documentation


iamrommel:
Ok. i'll remove this pull request, i added another one from v14 branch, which follows the latest useLocalStorage with added remove on the last item of array

PR #979, opened 2020-02-16
--escape double quote in lint script to properly find the correct fil…

iamrommel:
# Description

sessionStorage and localStorage should have similar behavior, their only difference is where the data is being stored and it's persistence which should be abstracted for this library.

## Type of change
- [ ] **Breaking change**, The signature/arguments follows the useLocalStorage, this is breaking changes, because the previous last argument is just a boolean "raw", which i change it to object "options"
- [ ] New feature, added the useStorage, as the base for sessionStorage, localStorage, or even  "whatEverStorage" that follows the signature of localStorage
- [ ] New feature, Added the sessionStorage unit test
- [ ] New feature, Update the docs that describe about the config changes, for raw, serializer, de-serializer
- [ ] Bug fix, Removed the single-quote in lint script, to properly find the correct files and folder, this is not related to the changes, but the lint command does not find the files with single qoute, i tried with bash and command terminals both does not work there (we can discuss more on this if there is a need to)

 
# Checklist
- [x] Add documentation

### DIFF
--- a/docs/useSessionStorage.md
+++ b/docs/useSessionStorage.md
@@ -16,6 +16,7 @@ const Demo = () => {
       <div>Value: {value}</div>
       <button onClick={() => setValue('bar')}>bar</button>
       <button onClick={() => setValue('baz')}>baz</button>
+        <button onClick={() => remove()}>Remove</button>
     </div>
   );
 };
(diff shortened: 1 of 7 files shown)


nspaeth:
Could you put this pull request into a separate feature branch as per the [Contributing Guide](https://github.com/streamich/react-use/blob/master/CONTRIBUTING.md)?

I would like to add some additional functionality, but it conflicts with the existing `v14.0` branch.

Issue #755, opened 2019-11-06
Remove warnings from `useDeepCompareEffect`


garrettmaring:
**Is your feature request related to a problem? Please describe.**

`useDeepCompareEffect` gives warnings when using it with primitive dependencies. This is frustrating for the common case where dependencies could either `undefined` or an `object` that you'd like to be deep-compared.

**Describe the solution you'd like**

Remove warnings from `useDeepCompareEffect` *or* provide a way to suppress warnings.

xobotyi:
Any example?

garrettmaring:
```typescript
import React from "react";
import { useDeepCompareEffect } from "react-use";

export const useObjectDidChange = <T extends any>(didChange: () => void, dep: T) => {
   useDeepCompareEffect(() => {
    didChange();
   }, [dep]);
}

function App = () => {
  const [obj, setObj] = useState<object | undefined>(undefined);
  useObjectDidChange(() => console.log("obj changed"), obj);
  
  return <div />;
}
```

This code will log a warning whenever `obj` is set to `undefined`.

garrettmaring:
I understand that the warning is there to help users. We should be using `useEffect` when the values of the dependency array are primitives. However, it's possible for the values of the array to change from primitive to non-primitive (as shown above). In that case, there should be no warning. It is correct to use `useDeepCompareEffect` if the value of the dependency is either a primitive or a non-primitive.

That warning is only valid if it is always primitive.

garrettmaring:
If the above case doesn't satisfy you, I'd at least add the option to hide warnings from `react-use`. Warnings are definitely useful and I appreciate that they're in `react-use`. However, they clutter the console and, in this case, I simply don't agree with the warning but would still like to continue using `react-use`.

streamich:
@garrettmaring I agree, I think the warnings for primitives in `useDeepCompare` (and shallow compare) should be removed.

streamich:
But the other warning about no dependencies should probably stay.

callmetushar123:
Is this issue open? Can anyone assign this to me?

shrutipandya3:
Can I work on this if still open?

PR #1004, opened 2020-02-25
feat: remove warning from `useDeepCompareEffect`

garrettmaring:
It is common for state to be in either a nil or non-nil state where the non-nil state is not a primitive. This PR removes the warning that is fired in such an event.

```
import React from "react";
import { useDeepCompareEffect } from "react-use";

export const useObjectDidChange = <T extends any>(didChange: () => void, dep: T) => {
   useDeepCompareEffect(() => {
    didChange();
   }, [dep]);
}

function App = () => {
  const [obj, setObj] = useState<object | undefined>(undefined);
  useObjectDidChange(() => console.log("obj changed"), obj);
  
  return <div />;
}
```

# Description

As discussed in #755, `useDeepCompareEffect` fires a warning for primitives being compared. Given that there are more complex state interactions where primitives such as `undefined` and `null` are acceptable, this warning should be removed.

## Type of change

# Checklist

- [x] Add documentation

### DIFF
--- a/src/useCustomCompareEffect.ts
+++ b/src/useCustomCompareEffect.ts
@@ -1,19 +1,11 @@
 import { DependencyList, EffectCallback, useEffect, useRef } from 'react';
 
-const isPrimitive = (val: any) => val !== Object(val);
-
 type DepsEqualFnType = (prevDeps: DependencyList, nextDeps: DependencyList) => boolean;
 
 const useCustomCompareEffect = (effect: EffectCallback, deps: DependencyList, depsEqual: DepsEqualFnType) => {
   if (process.env.NODE_ENV !== 'production') {
     if (!(deps instanceof Array) || !deps.length) {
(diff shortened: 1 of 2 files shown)


jamesmstone:
I hate to be that "_+1_ _person_"  - but...
...I would find this PR really helpful. 


Is there a particular reason it hasn't been merged,  @streamich , I see in #755 you seem to be open to its change.

Aside: Thanks for a really **use**ful lib! 

Cheers,
James

RIP21:
It's been a long time.
Can we please have this one shipped? :)  As it's a pretty small, but a nice change.

I've been maintaining[ `react-hanger`](https://github.com/kitze/react-hanger) for some time and kinda changed to be using `react-use` mostly recently as it's one of the most complete collections of all kinds of hooks on the internet :) 

I'll be happy to maintain this one too.

PR #1014, opened 2020-02-28
Added documentation for useTitle options

aej11a:
# Description

Small documentation addition to useTitle

### DIFF
--- a/docs/useTitle.md
+++ b/docs/useTitle.md
@@ -14,3 +14,16 @@ const Demo = () => {
   return null;
 };
 ```
+## Options
+
+If `restoreOnUnmount` is true, when the parent component is unmounted, the page title will be reverted to the title before the hook was called.
+
+```jsx
+import {useTitle} from 'react-use';
+
+const Demo = () => {
+  useTitle('Hello world!', { restoreOnUnmount: true });
+
+  return null;
+};
+```


dizel3d:
LGTM
Honestly, I surprised. Why we need `restoreOnUnmount` as option? Why not to do it always?

zsh2401:
> LGTM
> Honestly, I surprised. Why we need `restoreOnUnmount` as option? Why not to do it always?
> I found [PR](https://github.com/streamich/react-use/pull/776) but didn't catch the reason. @zsh2401 @streamich
If we do n’t make it optional,it might broke some systems down that depends on it.

zsh2401:
> LGTM
> Honestly, I surprised. Why we need `restoreOnUnmount` as option? Why not to do it always?
> I found [PR](https://github.com/streamich/react-use/pull/776) but didn't catch the reason. @zsh2401 @streamich

If we do n’t make it optional,it might broke some systems down that depends on it.

dizel3d:
> If we do n’t make it optional,it might broke some systems down that depends on it.

@zsh2401, compare:
```ts
useTitle('myTitle', { restoreOnUnmount: false });
useMount(() => document.title = 'myTitle'); // do the same
```
Was it really worth to make a breaking change in `useTitle`?

PR #1021, opened 2020-03-04
feat(createGlobalState)!: Align API with React.useState

nspaeth:
# Description
Adds support for functional updates to align with the React.useState
API. Type definitions are updated to match useState.

This might break existing code that expects functions to be held in the global
state, rather than applied as an update.

This change may also cause new type errors.

## Type of change

# Checklist
- [x] Add documentation


xobotyi:
There is a function and types in utils for [native-like state management](https://github.com/streamich/react-use/blob/master/src/util/resolveHookState.ts)

nspaeth:
> There is a function and types in utils for [native-like state management](https://github.com/streamich/react-use/blob/master/src/util/resolveHookState.ts) [](/streamich/react-use/blob/HEAD@{2020-03-05T06:14:29Z}/src/util/resolveHookState.ts)

I'll update the PR tomorrow to use those types.

nspaeth:
@xobotyi Updated.

nspaeth:
@xobotyi Is anything else needed in this PR, or in #1021 or #979 ?

src/createGlobalState.ts
```diff
@@ -1,19 +1,38 @@
...
+    setState(newState: HookState<S>) {
+      store.state = resolveHookState(newState);
       store.setters.forEach(setter => setter(store.state));
     },
     setters: [],
   };
 
-  return (): [S | undefined, (state: S) => void] => {
-    const [globalState, stateSetter] = useState<S | undefined>(store.state);
+  // unlike in useState, here initialState is optional because there will always
```
mbest, line 25: `useState` does not require an initial state if you accept `undefined` as a possible value.

nspaeth: I think my comment wasn't quite clear. I'm contrasting `useState` with the `useGlobalState` created by this function. For example, if you have a state
```ts
const [myDefinedState, setMyDefinedState] = useState<string>(INITIAL_STRING)
```
`INITIAL_STRING` is required, because the type does not include `undefined` in its definition.
In contrast:
```ts
const useMyDefinedState = createGlobalState<string>(DEFAULT_STRING)
....
const [ myDefinedState, setMyDefinedState ] = useMyDefinedState()
```
The initializer for `useMyDefinedState` is not required because it was previously supplied by the default param passed to `createGlobalState`.

mbest: React includes two definitions of `useState`. The one without a parameter automatically adds `undefined` to the value type:

```ts
function useState<S = undefined>(): [S | undefined, Dispatch<SetStateAction<S | undefined>>];
```

So if you have `const [myDefinedState, setMyDefinedState] = useState<string>()`, the type of `myDefinedState` becomes `string | undefined`.

Of course, this is still different from your suggested `useGlobalState`, which doesn't change the value type.

nspaeth: Oh, I didn't realize react did that. In that case, it sounds like just removing that comment will address your review?

mbest: Yes, please fix the comment.

tests/createGlobalState.test.ts
```diff
@@ -6,16 +6,68 @@ describe('useGlobalState', () => {
...
-  it('both components should be updated', () => {
+  it('both components should be updated directly', () => {
     const useGlobalValue = createGlobalState(0);
     const { result: result1 } = renderHook(() => useGlobalValue());
     const { result: result2 } = renderHook(() => useGlobalValue());
-    expect(result1.current[0] === 0);
-    expect(result2.current[0] === 0);
+    let [ state1, setState1 ] = result1.current;
+    let [ state2 ] = result2.current;
+    expect(state1 === 0);
```
mbest, line 15: I don't think this does anything. You need to follow `expect` with `to...`. So this should be `expect(state1).toBe(0)`.

src/createGlobalState.ts
nspaeth: Fixed

tests/createGlobalState.test.ts
nspaeth: Fixed

nspaeth:
I've updated this PR to address comments, and to account for a comment made [here](https://github.com/streamich/react-use/pull/1060#issuecomment-615346445), superseding #1060
 
@xobotyi @streamich 
Is there anything more I can do to help get this or #1025 merged?

xobotyi:
Watched it briefly, looks good for me.
@streamich merge is up to you.

nspaeth:
FYI, my local branch has a couple of type improvements that I plan to commit tomorrow.

ThomasAribart:
@nspaeth @xobotyi Is this pull request still planned to be merged ?

nspaeth:
I never got around to pushing some type fixes. I'll try to get to that
tonight.

LRNZ09:
Any update on this?

PR #1060, opened 2020-03-22
fix: SSR useEffect warning in createGlobalState hook

### LINKED ISSUE #956: `createGlobalState` should use `useIsomorphicLayoutEffect`
currently the hook generated by the `createGlobalState` call will warn server side, due to the `useLayoutEffect` call.

blnk-space:
#662 Description

createGlobalState uses useLayoutEffect which can't run on the server due to there being no DOM. This results in a warning that I have fixed by using useIsomorphicLayoutEffect.
This PR fixes #956 

## Type of change

# Checklist
- [x] Add documentation

### DIFF
--- a/src/createGlobalState.ts
+++ b/src/createGlobalState.ts
@@ -1,6 +1,7 @@
 /* eslint-disable */
-import { useLayoutEffect, useState } from 'react';
+import { useState } from 'react';
 import useEffectOnce from './useEffectOnce';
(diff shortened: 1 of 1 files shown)


mbest:
`createGlobalState` shouldn't be using either of these. The subscription should happen synchronously to avoid race conditions. Here's a synchronous `useMount` version we could use:

```ts
function useMount(fn: () => void): void {
    const isFirst= useFirstMountState();
    if (isFirst) {
        fn();
    }
}
```

PR #1119, opened 2020-04-09
Make sure hook won't break when using SSR

daphnesmit:
# Description

- Add fallbacks for window and document when your app is using SSR, for example a NextJS app 
- Remove boolean typecheck, because typescript says its already derived from default value
- Remove eslint-disable, don't know why it was there?
- Add state to return cleanup, because now linter want me to add that

## Type of change

### DIFF
--- a/src/useIdle.ts
+++ b/src/useIdle.ts
@@ -1,4 +1,3 @@
-/* eslint-disable */
 import { useEffect, useState } from 'react';
 import { throttle } from 'throttle-debounce';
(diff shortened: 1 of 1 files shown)


daphnesmit:
Is it possible to merge this ?

FlickerLogicalStack:
Does it make sense to do these checks if the `useEffect`'s callback is not executing on SSR?

daphnesmit:
I have no idea how I tested it

PR #1127, opened 2020-04-11
feat: useValue, useFunction

dizel3d:
# Description

## Type of change

# Checklist
- [x] Add documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,6 +155,8 @@
     <br/>
 - [**Miscellaneous**]()
   - [`useEnsuredForwardedRef`](./docs/useEnsuredForwardedRef.md) and [`ensuredForwardRef`](./docs/useEnsuredForwardedRef.md) &mdash; use a React.forwardedRef safely. [![][img-demo]](https://streamich.github.io/react-use/?path=/story/state-useensuredforwardedref--demo)
+  - [`useFunction`](./docs/useFunction.md) &mdash; mutable alternative to `useCallback`. [![][img-demo]](https://streamich.github.io/react-use/?path=/story/miscellaneous-usefunction--demo)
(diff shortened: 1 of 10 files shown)


mbest:
`useCallback` doesn't require a dependency list. Your `useFunction` isn't needed.

dizel3d:
> `useCallback` doesn't require a dependency list. Your `useFunction` isn't needed.

@mbest The difference is `useCallback` returns a new function on each render. It produces unnecessary rerenders in dependent children.

Here's an [example](https://jsfiddle.net/dizel3d/k6e0jsgv/27/):
```jsx
const Button = React.memo(({ onClick }) => {
    const count = React.useRef(0);
    count.current += 1;
    return (
      <div>
        <button onClick={onClick}>Button</button> rendered {count.current} times.
      </div>
    );
});

const App = () => {
    const [secs, setSecs] = React.useState(0);
    React.useEffect(() => {
        let time = 0;
        setInterval(() => setSecs(time++), 300);
    }, []);
    const onClick = React.useCallback(() => alert(secs));
    return (
        <Button onClick={onClick} />
    );
};

ReactDOM.render(<App />, document.querySelector('#root'));
```

If you replace `useCallback` with `useFunction`, `Button` will be rendered just once.

mbest:
What I meant to say is that you can pass an empty dependency array to [useCallback](https://reactjs.org/docs/hooks-reference.html#usecallback) to get the same functionality.

dizel3d:
> What I meant to say is that you can pass an empty dependency array to [useCallback](https://reactjs.org/docs/hooks-reference.html#usecallback) to get the same functionality as your `useFunction`.

No, it's not true.
Try to modify the example above. Button will always show 0 on click.

mbest:
Thanks for the clarification and the examples. I understand now. `useCallback(fn, [])` doesn't cause re-renders, but also only uses the original function reference. Your `useFunction` also doesn't cause re-renders, but always uses the latest function reference.

dizel3d:
@xobotyi is react-use still alive?

PR #1159, opened 2020-04-22
Improve TS typings on useAsyncFn

kolodny:
# Description

This PR improves overall TS typings on `useAsyncFn` as well as fix this particular issue:

```ts
const [, fn] = useAsyncFn((num: number) => Promise.resolve(num + 1));;
fn() // This should error, but doesn't with the current typings.
```

## Type of change

# Checklist
- [ ] Add documentation (Should be transparent to user)
- [ ] Cover changes with tests (No dtslint or tsd tests yet to test errors 😢)

### DIFF
--- a/src/useAsync.ts
+++ b/src/useAsync.ts
@@ -3,7 +3,7 @@ import useAsyncFn from './useAsyncFn';
 
 export { AsyncState, AsyncFn } from './useAsyncFn';
 
-export default function useAsync<Result = any>(
+export default function useAsync<Result>(
   fn: () => Promise<Result>,
   deps: DependencyList = []
 ) {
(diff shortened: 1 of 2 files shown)


src/useAsync.ts
```diff
@@ -3,11 +3,11 @@ import useAsyncFn from './useAsyncFn';
 
 export { AsyncState, AsyncFn } from './useAsyncFn';
 
-export default function useAsync<Result = any, Args extends any[] = any[]>(
```
FlickerLogicalStack, line 6: maybe `Result` can be with default type like `unknown`? Same as `Promise`'s default

kolodny: Done, it now uses the default `unknown` instead of of `any`. This value will be inferred to something else if the return type of the Promise can be inferred (which is almost always can)

kolodny:
Friendly ping on this PR 😃

PR #1197, opened 2020-05-06
fix: allow only for client

### LINKED ISSUE #1195: useHash is not SSR compatible

Error `window is undefined` is thrown`

Just use `useHash()` on SSR project.

If there isn't `window` object use some default values.

sungwoo91:
Fix #1195
 
# Description

Add line to check if it is a client.

## Type of change

# Checklist
- [ ] Add documentation


whinc:
I agree @streamich  . I think we should keep the client side and  server  side logic individual for better maintainability.

sungwoo91:
@streamich @whinc 
Thanks for your help! I'll fix it soon.

sungwoo91:
@streamich 
I changed the code with your advice. sorry for late!

kachkaev:
This solution may lead to issues with DOM rehydration. For example, a Next.js app expects the Client DOM after the first render to match the DOM that SSR phase produced. So in order to make things fully SSR compatible, we'll need to start ignore `window.location.hash` on the client during the first render and then sync up in `useEffect.` This is will solve SSR rehydration but will probably received not so well by the client-only apps. So how about introducing a new hook instead of 'fixing' an existing one?

mmahalwy:
any update to this? :)

PR #1203, opened 2020-05-10
feat: add useLatest hook

ClementParis016:
# Description

When working with React I regularly hit [the "stale closure" caveat of Hooks](https://reactjs.org/docs/hooks-faq.html#why-am-i-seeing-stale-props-or-state-inside-my-function) while trying to get the latest value of a prop in some asynchronous callback and I end up rewriting the same small snippet of code to handle that every time.

So, given this package already has `usePrevious()`, I thought I could suggest to add `useLatest` :)

Any feedback would be more than welcome! I personally don't like much the fact that it returns a `MutableRefObject` but that's the only way to get it to work because returning the `.current` property of the ref would make it behave exactly as `usePrevious()`.

PS: I didn't dig too much and mostly copy/pasted then adapted the test and story from `usePrevious()`, so please let me know if I did anything wrong here.

## Type of change

# Checklist
- [x] Add documentation


src/useLatest.ts
```diff
@@ -0,0 +1,13 @@
...
+
+const useLatest = <T>(value: T): MutableRefObject<T> => {
+  const latest = useRef<T>(value);
+
+  useEffect(() => {
+    latest.current = value;
+  }, [value]);
+
+  return latest;
+};
```
streamich, line 11: Do we need to put this into `useEffect`? Should it be out of the `useEffect`, like so:

```suggestion
const useLatest = <T>(value: T): MutableRefObject<T> => {
  const latest = useRef<T>(value);
  latest.current = value;
  return latest;
};
```

My guess is it should still work and be more performant, and new value updates will happen instantly instead of when scheduled by `useEffect`.

ClementParis016: Indeed it works, I did the change in 5488f5e.

My mileage may vary but I thought that it was a bit of an anti-pattern to perform side-effects at render in React?

ClementParis016:
@streamich I see you introduced a `useLatest` in #1217 which surfaces as a conflict here. Do you want to proceed with this PR or follow-up next on your changes?

src/useLatest.ts
chrisk-7777: @ClementParis016 @streamich I understand this was merged long ago, but there is little response in the issue tracker at the moment. 
Refs should not be mutated outside of useEffect or an event handler as they will not be compatible with concurrent mode.

xobotyi: @chrisk-7777 as concurrent mode is still experimental for a while i dont believe it will release any time soon - there is no sense to try adjust to it, as literally everything can change.

chrisk-7777: Even if you ignore concurrent mode, its still generally recommended to avoid by the Hooks FAQ: 
https://reactjs.org/docs/hooks-faq.html#is-there-something-like-instance-variables

> Unless you’re doing lazy initialization, avoid setting refs during rendering — this can lead to surprising behavior. Instead, typically you want to modify refs in event handlers and effects

But up to you guys I guess 🤷

PR #1221, opened 2020-05-18
fix: 🐛 correct useMeasure typings

Svish:
When upgrading to v15, I discovered some issues with `useMeasure` forcing me to downgrade to v14. Hopefully these changes fixes those issues...

* It was no longer possible to define the ref type, meaning components requiring a certain type of ref would reject it. Fixed this by making the function generic again.
* Previously you _had to_ define the ref type, while in v15 you _couldn't_. Fixed this by giving the generic type a default type of `HTMLElement`.
* The implicit typing of `useMeasureMock` combined with the ternary default export statement made the imported type of `useMeasure` very messy. Fixed this by making the type of `useMeasureMock` explicitly match `useMeasure`.

# Description

## Type of change

# Checklist

### DIFF
--- a/src/useMeasure.ts
+++ b/src/useMeasure.ts
@@ -6,8 +6,8 @@ export type UseMeasureRect = Pick<
   DOMRectReadOnly,
   'x' | 'y' | 'top' | 'left' | 'right' | 'bottom' | 'height' | 'width'
 >;
-export type UseMeasureRef = (element: HTMLElement) => void;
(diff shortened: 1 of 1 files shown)


Svish:
Most of the Checklist is irrelevant for this change. Maybe consider making a different Checklist for new features and bug fixes? 🤔

coffeedoughnuts:
#1222

PR #1279, opened 2020-06-06
feat: improve implementation of useFirstMountState() hook

LeoooY:
# Description

Improve implementation of useFirstMountState() hook.
I think use `useEffect` can be more efficient when frequently re-rendering 

## Type of change

# Checklist
- [ ] Add documentation

### DIFF
--- a/src/useFirstMountState.ts
+++ b/src/useFirstMountState.ts
@@ -1,13 +1,11 @@
-import { useRef } from 'react';
+import { useRef, useEffect } from 'react';
 
 export function useFirstMountState(): boolean {
   const isFirst = useRef(true);
 
-  if (isFirst.current) {
-    isFirst.current = false;
-
-    return true;
-  }
+  useEffect(() => {
+    isFirst.current = false
+  }, [])
 
   return isFirst.current;
 }


xobotyi:
It won't be effective because instead of regular comparison on each render your impelemtation will create new function and call `useEffect` hook.

LeoooY:
Since the deps of `useEffect` is `[]`, the effect will only run once when the component mounts and never run again. The old way will assign to `isFirst.current` on every render, I think it`s unnecessary.

xobotyi:
`[]` deps only means that provided function will be called once on first mount, but provided function itself will be dynamically created on each render.

xobotyi:
So if we will list the step-by-step execution log:

originally:
- get ref with default to `true`
- check `isFirst.current` is truthy
  - if truthy - assign `false`
  - return `true`
- return `isFirst.current` if falsy (actually it could be simplifier to `return false;`)

this PR:
- get ref with default to `true`
- create anonimous function
- create empty array
- pass them to `useEffect`
  - *inside `useEffect`*
    - check if dependencies contents changed
      - if changed - call provided function
- return `isFirst.current`

LeoooY:
Thanks for your patient explanation, I think I get the point.

PR #1295, opened 2020-06-15
feat(usePrevious): reworked the hook, now it is more memory-efficient.

xobotyi:
# Description

Better to use two-variables exchange than `useEffect` with new
function on each render.

## Type of change

# Checklist
- [x] Add documentation

### DIFF
--- a/src/usePrevious.ts
+++ b/src/usePrevious.ts
@@ -1,13 +1,11 @@
-import { useEffect, useRef } from 'react';
+import { useRef } from 'react';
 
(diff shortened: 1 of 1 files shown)


dcporter:
Hi, sorry to roll in like this but this PR was a mistake. The `useEffect` is necessary in cases where React renders multiple times before committing — something that it's within its rights to do, and something that the excellent `<React.StrictMode>` does on purpose to root out buggy side-effects like this.

PR #1416, opened 2020-08-01
fix(usePermission): avoid undefined exception

### LINKED ISSUE #1415: usePermission throws exception in iOS Safari

When testing on BrowserStack with a physical iPhone 11 in Safari, `navigator.permissions.query` is throwing an exception within `usePermission`.

image

Sign up for BrowserStack (free) and open an app that uses `usePermission` in Safari.

`usePermission` should probably no-op if the Permissions API is not supported. Alternatively, for each kind of permission, it could try and support what Safari is able to support (notifications, mostly).

See Google polyfill for an example: https://github.com/chromium/permissions.request

**Workaround**

I haven't found one yet since Rule of Hooks states you can't conditionally execute a hook. Polyfilling the APIs that I need (notifications) may be the only alternative that is compatible.

**A little about versions:**  
- _OS_: iOS
- _Browser (vendor and version)_: Safari v13
- _React_: 16.13
- _`react-use`_: 15.1.1

kamranayub:
# Description

Closes #1415 

Avoid throwing an undefined exception at runtime due to missing Permissions API. It would be preferable to wrap/abstract permissions for specific APIs in browsers like Safari but that could be out-of-scope for this hook. At the least, we just want to avoid bombing if the API is missing.

## Screenshots

From iPhone 11 in BrowserStack, Storybook correctly displays a blank string in Safari.

image

Previous code in Storybook:

image

## Type of change

# Checklist
- [x] Add documentation


src/usePermission.ts
```diff
@@ -30,13 +30,16 @@ const usePermission = (permissionDesc: PermissionDesc): State => {
...
+    // Permissions API is not available in every browser
+    if (navigator.permissions && navigator.permissions.query) {
+      navigator.permissions
+        .query(permissionDesc)
+        .then((status) => {
+          permissionStatus = status;
+          changeState();
+        })
+        .catch(noop);
+    }
```
EL-shadow, line 42: If the API is not supported, then we will not know about it. How about making a special status `unavailable` here?
```
} else {
    permissionStatus = 'unavailable';
    changeState();
}
```

EL-shadow: I tried my example and understand that this is not the best solution. 
Better this way
```
} else {
    setState('unavailable');
}
```

src/usePermission.ts
```diff
@@ -30,13 +30,16 @@ const usePermission = (permissionDesc: PermissionDesc): State => {
...
   useEffect(() => {
-    navigator.permissions
-      .query(permissionDesc)
-      .then(status => {
-        permissionStatus = status;
-        changeState();
-      })
-      .catch(noop);
+    // Permissions API is not available in every browser
+    if (navigator.permissions && navigator.permissions.query) {
```
EL-shadow, line 34: Can be simplified `navigator.permissions?.query`

kamranayub: Looks like this was already updated to

```
if ("permissions" in navigator) {
```

kamranayub:
oh no

kamranayub:
This issue is already fixed in `master` so I'm just going to go ahead and close this 🏃

PR #1424, opened 2020-08-05
fix(useUnmountPromise): Keep working after initial render

mogelbrod:
# Description

The function returned from `useUnmountPromise` currently doesn't resolve promises passed after the initial render. This happens due to the `useEffect` mount check missing a dependency array, which causes the hook to consider the component unmounted right after mounting.

## Type of change

# Checklist
- [x] Comment the code, particularly in hard-to-understand areas - not necessary in this case
- [x] Add documentation - not necessary in this case
- [x] Add hook's story at Storybook - not necessary in this case

### DIFF
--- a/tests/useUnmountPromise.test.ts
+++ b/tests/useUnmountPromise.test.ts
@@ -15,10 +15,14 @@ describe('useUnmountPromise', () => {
   it('when component is mounted function should resolve with wrapped promises', async () => {
     const hook = renderHook(() => useUnmountPromise());
 
-    const mounted = hook.result.current;
-    const res = await mounted(new Promise(r => setTimeout(() => r(25), 10)));
+    const res1 = await hook.result.current(new Promise(r => setTimeout(() => r(25), 10)));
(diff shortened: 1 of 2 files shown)


tests/useUnmountPromise.test.ts
```diff
@@ -15,10 +15,14 @@ describe('useUnmountPromise', () => {
...
     const hook = renderHook(() => useUnmountPromise());
 
-    const mounted = hook.result.current;
-    const res = await mounted(new Promise(r => setTimeout(() => r(25), 10)));
+    const res1 = await hook.result.current(new Promise(r => setTimeout(() => r(25), 10)));
+    expect(res1).toBe(25);
 
-    expect(res).toBe(25);
+    // FIXME: Ensure the useEffect() hook has executed before triggering next;
+    // using hook.waitForNextUpdate() doesn't work
```
mogelbrod, line 22: I've been unable to update the tests to make the current implementation fail, if anyone could help out that would be much appreciated!

PR #1435, opened 2020-08-12
useHash: add replaceInHistory parameter to setter

### LINKED ISSUE #1434: useHash: history replace (instead of push)
### Current implementation

When setting the hash through the `useHash` hook, the new hash is [pushed to the history](https://github.com/streamich/react-use/blob/master/src/useHash.ts#L22):
```ts
window.location.hash = newHash
```

### Feature request

I would like to be able to replace the current page in history with the new hash instead, by doing something like the following:

```ts
window.location.replace(window.location.href.replace(/#.*/, newHash))
```

> Note that `newHash` MUST start with `#` to use the above code otherwise the path is changed

### Motivation

If we use `useHash` as a `useState<string>` to fully-controll an `<input>`, it works quite well:

```ts
const Example: FC = () => {
  const [hash, setHash] = useHash();
  return <input value={hash.substr(1)} onChange={e => setHash(e.target.value)} />;
};
```
The only issue is that the history keeps track of every change, so if we click on the _previous_ button of the browser, it undoes the last key press. This makes it really hard to go to the previous page.

### Proposal

I suggest that the setter function returned by the hook can have an extra `replaceInHistory` boolean parameter. That parameter would be optional and would default to `false` for backward compatibility.

```ts
(newHash: string, replaceInHistory?: boolean) => void
```

Thoughts? @whinc

Rogdham:
Fixes #1434

# Description

As described in #1434 and validated by @whinc the author of `useHash`:

 - Adds a `replaceInHistory` parameter to the setter; when true, the current history entry is replaced with the new hash instead of creating a new entry.

```ts
(newHash: string, replaceInHistory?: boolean) => void
```

 - Makes it clear that it is optional for the `newHash` parameter to start with `#`
 - If `newHash` parameter does not start with `#`, avoids setting the hash if the hash did not change (e.g. hash was `#foobar` and setting it to `foobar`)

:arrow_forward: all these changes should be backward compatible with current implementation

## Type of change

# Checklist
- [X] Add documentation
- [X] Add hook's story at Storybook


dcporter:
I know I’m late to this party, but can I suggest an options object instead of a Boolean? That documents itself (avoids the so-called “Boolean trap”) and allows for expanding options in the future, which seems likely in the case of this API. Check out how the browser `addEventListener` API supports a bool (for capture I think) and also an options object, added later when they realized they had more options they wanted to support 😄

PR #1438, opened 2020-08-13
fix(useLocalStorage): No undefined types with defined initialState

Svish:
# Description

The type of the returned value from `useLocalStorage` is currently typed as `T | undefined`, but when providing an `initialValue`, it should be unnecessary to check for `undefined`, since, if it is, you should get the `initialValue` of type `T` instead.

To solve this I have added two TypeScript overloads for `useLocalStorage`:

1. The case when you provide `initialValue`, value is `T`
2. The case where you don't, value is `T | undefined`.

Additionally:

* I changed the type of the setter parameter type to not allow `undefined`, since that's a no-op anyways.
* Changed the `remove` callback to set state back to `initialValue` instead of `undefined`.
* Wrote a test for `remove` as I couldn't find any existing ones.

I have tested to install this package locally (via `yarn pack`) to check if the types look correct in a project using it, and as far as I can see, it does.

One thing I'm a bit uncertain of, is if the `set` callback should allow `undefined`, when `initialValue` is `undefined`. It would make sense, but since that's a no-op anyways, it kind of doesn't. So my conclusion is that `set` should never allow `undefined`, unless `undefined` actually does something. It could for example be a second way to remove/unset the value (i.e. do exactly the same as `remove`), but then you might as well just use `remove`. So, yeah...

## Type of change

The breaking changes are:

* The typings of the `useLocalStorage` hook and callbacks.
* That `remove` now sets state to `initialValue` rather than `undefined`

# Checklist
- [x] Add documentation


Svish:
Example of why the current types are annoying:

image

Here, there's no reason why `value` should be `undefined`, since I provide the hook with an `initualValue` that is guaranteed to be a `boolean`.

gongjoonamu:
I legit need this right now lol merge please! I'm having to sprinkle in a non-null assertion #ugly

Svish:
@gongjoonamu If you find yourself sprinkling many of those non-null assertions, if you haven't already done similar, here's a temporary "workaround" to keep the #ugliness is one place. So when this is released, you can just delete this, and "redirect" the imports back to the `react-use` one.

```ts
import { Dispatch, SetStateAction } from 'react';
import { useLocalStorage as useLS } from 'react-use';

// TODO: Remove when https://github.com/streamich/react-use/pull/1438 (hopefully) gets released
export default function useLocalStorage<T>(
  key: string,
  initialValue: T
): [T, Dispatch<SetStateAction<T>>, () => void] {
  return useLS<T>(key, initialValue) as [T, Dispatch<SetStateAction<T>>, () => void];
}
```

Svish:
Added another commit now, which replaces the current `useSessionStorage` with a copy of `useLocalStorage` (with storage replaced) because of the following:

* It had the same typing issue as `useLocalStorage`
* It did not have the same features, like custom serializer and such
* It did not have any tests at all

Svish:
Would be nice if someone could review this soon. 🙏🚀

Svish:
Realized I had another PR with a fix related to `useLocalStorage` too (#1448), so decided to move that commit here and apply the fix for `useSessionStorage` as well.

I also fixed some eslint warnings, and added some tests for the serializer & deserializer options which were previously untested.

Sjiep:
Looks like this MR also will fix the return type of `useSessionStorage`? 

Currently it is returning `[T, (value: T) => void]`, and I am running into the problem that Typescript is complaining that I can't use the functional update for the setState function.

Whereas is should be, `[T, Dispatch<SetStateAction<T>>]`, or since you use the copy of useLocalStorage `[T, Dispatch<SetStateAction<T>>, () => void]`

Svish:
> Looks like this MR also will fix the return type of `useSessionStorage`?
> 
> Currently it is returning `[T, (value: T) => void]`, and I am running into the problem that Typescript is complaining that I can't use the functional update for the setState function.
> 
> Whereas is should be, `[T, Dispatch<SetStateAction<T>>]`, or since you use the copy of useLocalStorage `[T, Dispatch<SetStateAction<T>>, () => void]`

Yeah, this PR makes them the same api and typings.

dantman:
Are there any issues with this PR?

I needed to switch a useLocalStorage using a custom (de)serializer with useSessionStorage and was disappointed to find that it was missing that functionality and I couldn't use it.

Svish:
Noticed some merge-conflicts had popped up here since last time, so I have rebased and fixed them all now.

makinde:
@streamich any chance you or another maintainer can look/merge?

sirjuan:
I closed my PR #1506 in favor of this one. 

I would use `useLatest` or similar approach for `serializer`/``deserializer` or even `set` itself to maintain referential stability similar to `React.useState`.

offirgolan:
Any update on getting this merged?

nikeee:
My current project would benefit from this. Any chance getting this merged?

isaacd8k:
Bump... Hoping this can get merged and released some time soon

SaburovBabur:
Any updates? I am having same issue here. `undefined` even though second param will be always defined

lmossman:
@streamich are you able to review this? It would help out a lot of us to get this merged in

tests/useSessionStorage.test.ts
```diff
@@ -0,0 +1,299 @@
...
+        return result.current;
+      })();
+      fail('hook should have thrown');
+    } catch (e) {
+      expect(String(e)).toMatch(/key may not be/i);
+    }
+  });
+
+  describe('remove', () => {
+    it('removes value from sessionStorage and restores initialValue', () => {
```
ahmetacer5, line 178: I think we need to distinguish between `remove` and `restore` here.

When the consumer calls the remove function:

1. Does the user wants to delete the session key from the storage?
2. Does the user wants to restore the session key to it's initial value?

I do not have answer for it therefore I wouldn't introduce this new feature yet.

PR #1448, opened 2020-08-18
fix(useLocalStorage): Fixed function updater providing stale state

Svish:
# Description

The state provided to the `setState` callback from `useLocalStorage` is stale, which for example means that the below toggle usage only works once, and after that the state is "stuck". I.e. calling the `toggle` callback will only work the first time.

```ts
export default function useToggle(key: string, initialValue: boolean): [boolean, () => void] {
  const [value, setValue] = useLocalStorage(key, initialValue)

  const toggleValue = useCallback(() => {
    // calling toggleValue() will only work once, because state passed to setValue is "stuck" on first value
    setValue((value) => !value)
  }, [setValue])

  return [value, toggleValue]
}
```

## Type of change

Although, it might be breaking if someone relies on the faulty stuck/stale state...

# Checklist
- [x] Add documentation

### DIFF
--- a/src/useLocalStorage.ts
+++ b/src/useLocalStorage.ts
@@ -69,7 +69,7 @@ const useLocalStorage = <T>(
         // localStorage can throw. Also JSON.stringify can throw.
       }
     },
-    [key, setState]
+    [key, state, setState]
   );
 
   const remove = useCallback(() => {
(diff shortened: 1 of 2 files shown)


Svish:
Moved this change into #1438 instead.

PR #1465, opened 2020-08-25
feat: new hooks useCallbackState

kylemocode:
# Description

React state hook that allow us to bring callback function in state setter, which can do some stuff base on the new state value, make asynchronous behavior with react state easier.
(I think this is a common idea, but no one seems to have implemented it in react-use yet)

## Type of change

# Checklist
- [X] Add documentation
- [X] Add hook's story at Storybook


dcporter:
This is clever — track what actually changes (and is actually committed to the actual DOM!) without manually wiring up one effect per state.

- It took me a while to understand what this does. I wonder if `useStateWithCallback` might be a clearer name? A bit longer but that's what autocomplete is for 😄
- Would it make sense to also allow setting a callback from the hook itself _also_, passed in as a second arg to the hook? Maybe it would be nice or maybe it would be overcomplicated

kylemocode:
> This is clever — track what actually changes (and is actually committed to the actual DOM!) without manually wiring up one effect per state.
> 
> * It took me a while to understand what this does. I wonder if `useStateWithCallback` might be a clearer name? A bit longer but that's what autocomplete is for 😄
> * Would it make sense to also allow setting a callback from the hook itself _also_, passed in as a second arg to the hook? Maybe it would be nice or maybe it would be overcomplicated

Hi, thanks for your reply. I think "useStateWithCallback" is indeed a better name, so I change the hook name in the lattest commit.
Regarding the second proposal, I want to keep the usage of useStateWithCallback hook similar to the original React.useState, except for callback function can be added to state setter. I think that passing a callback as second argument  will make more confuse, maybe it doesn't make sense in this case.

Really appreciate your reply 😊

PR #1334, opened 2020-07-01
fix(useLocalStorage): support partial updates

signed:
# Description

Support for partial updates in useLocalStorage

## Type of change

- [x] Bug fix _(non-breaking change which fixes an issue)_
- [ ] New feature _(non-breaking change which adds functionality)_
- [ ] **Breaking change** _(fix or feature that would cause existing functionality to not work as before)_

# Checklist
- [x] Read the [Contributing Guide](https://github.com/streamich/react-use/blob/master/CONTRIBUTING.md)
- [x] Perform a code self-review
- [x] Comment the code, particularly in hard-to-understand areas
- [-] Add documentation
- [-] Add hook's story at Storybook
- [x] Cover changes with tests
- [x] Ensure the test suite passes (`yarn test`)
- [x] Provide 100% tests coverage
- [x] Make sure code lints (`yarn lint`). Fix it with `yarn lint:fix` in case of failure.
- [x] Make sure types are fine (`yarn lint:types`).

PR #1506, opened 2020-09-14
fix: useLocalStorage setState should use latest state in updater cb

### LINKED ISSUE #1379: useLocalStorage's setValue(value => ...) passes the initial value instead of the current value

If you use the function form of useLocalStorage's setValue the callback receives the initial value instead of the current value.

**Cause of the issue**

This part of the code passes `state` to the callback. However this comes from the closure and because `state` is not part of the deps array the `set` function is created on initial render and a new callback is not created when state changes.

https://github.com/streamich/react-use/blob/edbb10580e40520cf0552e089e6929436028b8fb/src/useLocalStorage.ts#L53

Potential fixes are A) passing `state` to the deps array (which will have the side effect of recreating `set` on every state change, which is different than how setState works; B) maybe reorganize the callback implementation to start with a function form of setState and do the serializing and localStorage inside the setState callback.

`setValue(value => ...)` should pass the current value to `value` the same way that `useState`'s setValue does.

sirjuan:
# Description

Fixes #1379 

Using `useLocalStorage`'s `setState` function with updater function, state passed to callback is stale (always `initialValue`). I would expects it to work like the way `useState`'s updater works. 

For instance below fails to correctly toggle state after first run:

```
const [state, setState] = useLocalStorage('foo', false);
const toggleState = setState(old => !old);
```

This pr fixes this by passing state to updater using `useLatest`-hook. 

This might affect some developers.

## Type of change

# Checklist
- [x] Add documentation

### DIFF
--- a/src/useLocalStorage.ts
+++ b/src/useLocalStorage.ts
@@ -56,16 +56,24 @@ const useLocalStorage = <T>(
     (valOrFunc) => {
       try {
         const newState = typeof valOrFunc === 'function' ? (valOrFunc as Function)(latestState.current) : valOrFunc;
-        if (typeof newState === 'undefined') return;
-        let value: string;
 
-        if (options)
-          if (options.raw)
-            if (typeof newState === 'string') value = newState;
-            else value = JSON.stringify(newState);
-          else if (options.serializer) value = options.serializer(newState);
-          else value = JSON.stringify(newState);
-        else value = JSON.stringify(newState);
+        if (newState === undefined) {
+          return;
+        }
+
+        const getValue = () => {
+          if (options?.raw) {
+            return typeof newState === 'string' ? newState : JSON.stringify(newState);
+          }
+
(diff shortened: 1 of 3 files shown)


makker:
Would be awesome to get this fix included in next release 👍

joltmode:
FYI, there are 3 more PR's with attempt to solve the same issue:
- https://github.com/streamich/react-use/pull/1350
- https://github.com/streamich/react-use/pull/1334
- https://github.com/streamich/react-use/pull/1272

sirjuan:
> FYI, there are 3 more PR's with attempt to solve the same issue:
> 
> * #1350
> * #1334
> * #1272

So it seems. I would accept them, although I prefer my solution since it retains `setState` reference when `state` changes like React's `useState`.

sirjuan:
Closed in favor of #1438

Issue #1227, opened 2020-05-18
[useMeasure] Ability to track a custom Ref


ashubham:
**Is your feature request related to a problem? Please describe.**
Currently `useMeasure` returns a `ref` which needs to be assigned a ref in the JSX.
This has 2 issues:
1. The `ref` is not a real `useRef` for eg. Does not have `ref.current` as a prop.
2. If I already have a component where I have a ref, I just want to track its size changes.

**Describe the solution you'd like**
An API of this shape:
const {width, height ... } = useMeasure({ref: currentRef});

**Describe alternatives you've considered**
I am using https://github.com/ZeeCoder/use-resize-observer ... in the absence of the above which I want to remove.

michelecocuccio:
+1 on that. I already have my `ref` so it will be useful to pass a Ref as an argument. This will also give the ability to use more than one `useMeasure` in the same component, as I need to track three different elements.

kybradeck:
I just ran in the same issue, would also love this!

dmitrij-borchuk:
Hey! How do you think, should we pass a real `ref` there or just an element?
```
const {width, height ... } = useMeasure({ ref: ref });
```
or
```
const {width, height ... } = useMeasure({ element: ref.current });
```
I am asking because `ref` is not used inside so we don't need it.

dmitrij-borchuk:
Looks like this proposal contains breaking backward compatibility. Should new PR implement it like that?
Currently, it returns an array where the first element is `ref`, but with passing `ref` as an argument, we don't need this returning `ref`. In case we need to save backward compatibility the interface will be strange:
```
const rect = useMeasure({ ref: elementRef })[1];
```
or
```
const { '1': stateArgs } = useMeasure({ ref: elementRef });
```

dmitrij-borchuk:
@ashubham @streamich What do you think?

joeshub:
what if you create and return a ref if no ref was passed in?

dmitrij-borchuk:
@joeshub Great idea for the JavaScript but how TypeScript types should look like in this case?

priley86:
+1 for supporting a passed ref

PoOw:
> @joeshub Great idea for the JavaScript but how TypeScript types should look like in this case?

Could be something like that ?

```
<E extends HTMLElement = HTMLElement>(element?: E): UseMeasureResult<E>
```

If the `element` parameter is passed, you use it to initiate the element state in the hook. This way, there is no breaking change.

[...]

PR #1516, opened 2020-09-17
Use measure track custom ref

dmitrij-borchuk:
# Description

According to this (https://github.com/streamich/react-use/issues/1227) issue, we need to be able to track custom Ref passed from outside of the hook. I have added overload for the case when we pass a custom hook as an argument. In this case, we return only the Element rect.

## Type of change

# Checklist
- [ ] Add documentation


macrozone:
can we have this reviewed and released?

adeguelle:
> can we have this reviewed and released?

+1
please

dzcpy:
You could use useMeasureDirty instead. It's already in this project but not exported.
Try
```js
import useMeasureDirty from 'react-use/lib/useMeasureDirty';
```

dmitrij-borchuk:
So, its a year since I have created a pull request, looks like the repository is dead

dzcpy:
> So, it's a year since I have created a pull request. Looks like the repository is dead

You can use ahooks instead

PR #1519, opened 2020-09-18
feat: add useBodySize hook

carbongo:
# Description

A new sensor hook _useBodySize_ added. It has various use-cases, e.g. when a whole page width and/or height are needed within Components. Personally I used it for a "smart" scroll-up button, which has a full stop on a certain point of the page.

## Type of change

# Checklist
- [x] Add documentation

### DIFF
--- a/src/useBodySize.ts
+++ b/src/useBodySize.ts
@@ -1,0 +1,34 @@
+import { useMemo } from 'react';
+import useIsomorphicLayoutEffect from './useIsomorphicLayoutEffect';
+import useRafState from './useRafState';
+import { isClient } from './util';
+
+const useBodySize = (initialWidth = Infinity, initialHeight = Infinity) => {
+  const [state, setState] = useRafState<{ width: number; height: number }>({
+    width: isClient ? document.body.clientWidth : initialWidth,
+    height: isClient ? document.body.clientHeight : initialHeight,
+  });
+
+  const observer = useMemo(
+    () =>
+      new (window as any).ResizeObserver((entries) => {
+        if (entries[0]) {
+          const { width, height } = entries[0].contentRect;
+          setState({ width, height });
+        }
+      }),
+    []
+  );
+
+  useIsomorphicLayoutEffect(() => {
+    if (!isClient) return;
+    observer.observe(document.body);
+    return () => {
+      observer.disconnect();
+    };
+  }, [document.body]);
+
+  return state;
+};
+
+export default useBodySize;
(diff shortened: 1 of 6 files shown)


src/useBodySize.ts
```diff
@@ -0,0 +1,34 @@
...
+
+const useBodySize = (initialWidth = Infinity, initialHeight = Infinity) => {
+  const [state, setState] = useRafState<{ width: number; height: number }>({
+    width: isClient ? document.body.clientWidth : initialWidth,
+    height: isClient ? document.body.clientHeight : initialHeight,
+  });
+
+  const observer = useMemo(
+    () =>
+      new (window as any).ResizeObserver((entries) => {
```
ngobach, line 14: Please correct me if I'm wrong.
`useMemo()` will also be called on SSR requests, won't it? If yes then this line of code might throw on SSR request.

ngobach:
I see your source code is very similar to an existing hook ([**useMeasure**](https://github.com/streamich/react-use/blob/master/src/useMeasure.ts)). It is even possible that your PR can be replaced by using this hook with `ref` pointed to `body`.
Also there is another hook called `useWindowSize` that might be what you wanted.

carbongo:
> I see your source code is very similar to an existing hook ([**useMeasure**](https://github.com/streamich/react-use/blob/master/src/useMeasure.ts)). It is even possible that your PR can be replaced by using this hook with `ref` pointed to `body`.
> Also there is another hook called `useWindowSize` that might be what you wanted.

There is a big difference with `useWindowSize`. `useWindowSize` calculates dimensions of window, but `useBodySize` shows you a size of a document (it's not static and not fully tied to a window size, especially if you have a lot of text, you can see the difference in Storybook). `useBodySize` meant to be called in high-level components, in case you don't want to change the root document for passing dimensions using `useMeasure` through a whole code, you can do so, but... well, considering the fact that I needed those dimensions only in one component, it could've been a bad choice.

ngobach:
Hi @carbongo ,
`useWindowSize` is not what you wanted, so please ignore that one.
About your second point, I think there is a small misunderstanding here, because I didn't mean that you will have to drill dimensions returned from `useBodySize` down to each components. `useMeasure` can be used right in those component with ref set to `document.body`.
If you still see `useBodySize` is an essential hook, I still think you should write it base on `useMeasure` hooks with ref set to `window.body` to avoid duplicated code.

Something like:

```js
function useBodySize() {
  const [setRef, rect] = useMeasure();
  useEffect(() => setRef(document.body));
  return [rect.width, rect.height];
}
```

PR #1564, opened 2020-10-12
Should only run effect when callback changes

christiaanwesterbeek:
# Description

As the documentation states, this is the interval hook based on [Dan Abramov's article on overreacted.io](https://overreacted.io/making-setinterval-declarative-with-react-hooks). But it sets the `savedCallback` ref again on every render. It's not that is breaks the code, but it's useless and wasteful. Should update only the `savedCallback` when the passed callback is a different one.

## Type of change

# Checklist
- [ ] Add documentation

### DIFF
--- a/src/useInterval.ts
+++ b/src/useInterval.ts
@@ -5,7 +5,7 @@ const useInterval = (callback: Function, delay?: number | null) => {
 
   useEffect(() => {
     savedCallback.current = callback;
-  });
+  }, [callback]);
 
   useEffect(() => {
     if (delay !== null) {


dcporter:
The `[callback]` dependency is required to keep the callback up to date, as a new function can be passed in every time.

christiaanwesterbeek:
My change is not preventing from the savedCallback to be updated when a new functions is passed. It merely prevents updating the savedCallback when the SAME function is passed.

dcporter:
(Yeah, sorry, I was posting in support of your PR. Not well though, pre-coffee 😖)

luisgurmendezMLabs:
Why does this even need to store the callback in a ref? Can't it just be passed to the last useEffect as a dep and used directly? @christiaanwesterbeek

luisgurmendezMLabs:
Following Dan's sandbox https://codesandbox.io/s/beautiful-benz-zuodw?file=/src/index.js, I think that this works perfectly aswell:

```
import React, { useState, useEffect, useRef } from "react";
import ReactDOM from "react-dom";

function Counter() {
  const [count, setCount] = useState(0);
  const [delay, setDelay] = useState(1000);
  const [isRunning, setIsRunning] = useState(true);
  const [sum, setSum] = useState(true);

  const intervalFn = sum? ()=>setCount(count + 1) : ()=> setCount(count - 1)
  useInterval(
    intervalFn,
    isRunning ? delay : null
  );

  function handleDelayChange(e) {
    setDelay(Number(e.target.value));
  }

  function handleIsRunningChange(e) {
    setIsRunning(e.target.checked);
  }

  function handleIsSumChange(e) {
    setSum(e.target.checked);
  }

  return (
    <>
      <h1>{count}</h1>
      <input
        type="checkbox"
        checked={isRunning}
        onChange={handleIsRunningChange}
      />{" "}
      Running
      <input
        type="checkbox"
        checked={sum}
        onChange={handleIsSumChange}
      />{" "}
      {sum? 'Sum' : 'Substract'} 
      <br />
      <input value={delay} onChange={handleDelayChange} />
    </>
  );
}

function useInterval(callback, delay) {
  // Set up the interval.
  useEffect(() => {
    function tick() {
      callback();
    }
    if (delay !== null) {
      let id = setInterval(tick, delay);
      return () => clearInterval(id);
    }
  }, [delay, callback]);
}

const rootElement = document.getElementById("root");
ReactDOM.render(<Counter />, rootElement);
```
Am I missing something?

christiaanwesterbeek:
The setInterval in your useInterval hook has now basically become a setTimeout. Why? Let's see what happens

Once the tick method is called, intervalFn is called, setCount is called and the Counter element will rerender.

When it rerenders a new intervalFn is created. Right? Because you didn't use useCallback (or useMemo) there.

So a new! callback method (the newly created intervalFn) is passed to useInterval, the useEffect receives it as one of its dependencies and detects it's a different callback method as the one before, therefore it will call its cleanup method (the thing the effect returns) and run its effect again, calling setInterval again.

Functionally this is not a problem, because your solution works. But more work is done by React than strictly required.

PR #1569, opened 2020-10-13
Fix useOrientation when used with SSR apps

edwinjoseph:
# Description
When using `useOrientation` in an SSR app it causes the error `window is undefined` to be thrown. Update the hook to use the utility `isClient` around the `useEffect` hook and move all references of the window to be inside that statement.

Fixes https://github.com/streamich/react-use/issues/1570
## Type of change

# Checklist

### DIFF
--- a/src/useOrientation.ts
+++ b/src/useOrientation.ts
@@ -1,5 +1,5 @@
 import { useEffect, useState } from 'react';
-import { off, on } from './util';
+import { off, on, isClient } from './util';
 
 export interface OrientationState {
   angle: number;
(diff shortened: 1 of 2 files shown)


dantman:
What's the point of the `isClient`? Last time I checked useEffect does not run on the server. So the first part, moving `window.screen` out of the hook and into the effect, should be enough to solve the bug.

edwinjoseph:
> What's the point of the `isClient`? Last time I checked useEffect does not run on the server. So the first part, moving `window.screen` out of the hook and into the effect, should be enough to solve the bug.

Yes, you are correct. I just used what was being done on useWindowSize - https://github.com/streamich/react-use/blob/master/src/useWindowSize.ts#L13. I can update this to remove the isClient check or I could move it before the useEffect to return early.

Issue #1582, opened 2020-10-19
useScroll - scroll direction? 


IrvingArmenta:
**Is your feature request related to a problem? Please describe.**
I'm working on a "chat" application right now which needs some fetching on scrolling while going "up", so my implementation needs to be aware of the direction scrolling, I think it would also help to some other kind of UI navigations/interactions.

**Describe the solution you'd like**
I would like the hook to return an object with something like:
```js
// ...current useScroll() implementation, where state is { x: number , y: number } plus the scrollDirection
 return { pos: state, dir: scrollDirection }
```
of course `dir` would be updated with a very good precision.

**Describe alternatives you've considered**
I'm trying to implement the scroll direction prop myself, but I couldn't get it to be completely responsive, I tried using the `usePrevious()` hook within the `useScroll()` hook to get what I wanted, but there is some "delay" when updating the scroll direction, I'm not sure how to completely get it "live".

szygendaborys:
I opened a pull request for this issue. #1634

PR #1635, opened 2020-11-12
Feature/use scroll direction

szygendaborys:
# Description

It is a solution for the issue #1582 .

Added additional field for the useScroll hook.
The field 'direction' contains fields 'horizontal' and 'vertical'. Initially, they both are set to 'none'. After scroll event it sets a proper direction state to both fields.

## Type of change

# Checklist
- [ ] Add documentation

### DIFF
--- a/src/useScroll.ts
+++ b/src/useScroll.ts
@@ -5,6 +5,10 @@ import useRafState from './useRafState';
 export interface State {
   x: number;
   y: number;
+  direction: {
+    horizontal: 'up' | 'down' | 'none';
+    vertical: 'left' | 'right' | 'none';
+  }
 }
 
 const useScroll = (ref: RefObject<HTMLElement>): State => {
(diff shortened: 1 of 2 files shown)


javiergarciagonzalez:
It'd be amazing to have this feature!

src/useScroll.ts
```diff
@@ -17,14 +21,35 @@ const useScroll = (ref: RefObject<HTMLElement>): State => {
...
+        }
+
         setState({
           x: ref.current.scrollLeft,
           y: ref.current.scrollTop,
+          direction: {
+            horizontal, vertical
+          }
         });
       }
```
fearphage, line 54: I have a suggestion to simplify and un-nest the code, exit early. I also aligned the formatting (trailing commas, space after `if`).

```suggestion
      if (!ref.current) {
        return;
      }
      
      const { scrollLeft, scrollTop } = ref.current;

      let horizontal: 'up' | 'down' | 'none' = 'none';
      let vertical: 'left' | 'right' | 'none' = 'none'; 
      
      if (state.y > scrollTop) horizontal = 'down';
      else if (state.y < scrollTop) horizontal = 'up';
  
      if (state.x > scrollLeft) vertical = 'left';
      else if (state.x < scrollLeft) vertical = 'right';
      
      setState({
        x: scrollLeft,
        y: scrollTop,
        direction: {
          horizontal, 
          vertical,
        },
      });
```

You could also rename `scrollLeft` and `scrollTop` to `y` and `x` respectively during destructuring, but I didn't make that change.

fearphage (changes requested):
Clean up suggestion

src/useScroll.ts
szygendaborys: Unfortunately, this package looks no longer to be maintained :(

PR #1684, opened 2020-12-06
fix(useStateWithHistory): support callback style setState

### LINKED ISSUE #1916: useStateWithHistory doesn't pass current value to callback style setState

Using `useStateWithHistory`'s callback style `setState` will receive no arguments.

```
const [state, setState] = useStateWithHistory(0);
setState(val => val + 1); // val is undefined
```

See example above.

Expected callback style `setState` to receive the current value as the first argument.

**A little about versions:**  
- _OS_: NA
- _Browser (vendor and version)_: All (Chrome)
- _React_: 17
- _`react-use`_: 13.3.0

QingqiShi:
# Description

Make `useStateWithHistory` support the callback style set state, e.g. `setState(current => current + 1)`. 

Currently, it's typed as if it does support it, but the `current` parameter is always `undefined`.

## Type of change

# Checklist
- [ ] Add documentation

Fixes #1916

### DIFF
--- a/src/useStateWithHistory.ts
+++ b/src/useStateWithHistory.ts
@@ -57,7 +57,7 @@ export function useStateWithHistory<S, I extends S>(
   const setState = useCallback(
     (newState: ResolvableHookState<S>): void => {
       innerSetState((currentState) => {
-        newState = resolveHookState(newState);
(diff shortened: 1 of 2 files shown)


xobotyi:
thanks

PR #1693, opened 2020-12-10
feat: add useSubscribedLocalStorage

xialvjun:
# Description

`useLocalStorage` with synced value of different tabs by `storage` event

## Type of change

# Checklist
- [ ] Add documentation

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -89,6 +89,7 @@ export { default as useSpeech } from './useSpeech';
 export { default as useStartTyping } from './useStartTyping';
 export { useStateWithHistory } from './useStateWithHistory';
 export { default as useStateList } from './useStateList';
+export { default as useSubscribedLocalStorage } from './useSubscribedLocalStorage';
 export { default as useThrottle } from './useThrottle';
 export { default as useThrottleFn } from './useThrottleFn';
 export { default as useTimeout } from './useTimeout';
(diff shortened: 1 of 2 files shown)


xialvjun:
or another name ? `useSyncedLocalStorage` ?

Svish (commented):
Would it maybe be better if this either managed to build upon `useLocalStorage` (although not sure that's possible), or if `useLocalStorage` just got another option (for example `subscribed: boolean`, or `sync: `boolean`)?

Svish:
Think it would be good if this worked the same as `useLocalStorage` with the fixes in my PR #1438, but not sure if it's possible to do that in a "wrapping way", so maybe best would be the latter, to add an option. 🤔

src/useLocalStorage.ts
```diff
@@ -47,6 +53,29 @@ const useLocalStorage = <T>(
...
+    useEffect(() => {
+      const callback = (e: StorageEvent) => {
+        if (e.key !== key) {
+          return;
+        }
+        if (e.newValue === null) {
+          return remove();
+        }
+        try {
+          setState(deserializer(e.newValue));
```
Svish, line 67: After having tried subscribing to storage events in a local variant of this hook, I've found this to be super unstable. It works fine for primitives (boolean, number, string), but not for arrays and objects which will cause an infinite update loop.

A subscribed/synced feature of this hook really needs to be tested thoroughly to make sure it's stable. Would probably need to do some sort of deep-equals check or something to prevent infinite updates.

PR #1724, opened 2020-12-31
Fix typo

nicgirault:


Svish (approved):
`1_000` is actually valid javascript, but personally find it a bit unnecessary and distracting to use that here.

nicgirault:
Oh! I didn't know that! Thanks

xobotyi:
Agree with @Svish, merging

PR #1745, opened 2021-01-12
fix: useMeasure type definitions for SVG

robinvw1:
Re-applied the changes of @thevtm from https://github.com/streamich/react-use/pull/479 to the `useMeasure` hook, because the Resize Observer API [also works](https://github.com/WICG/resize-observer/issues/9#issuecomment-228181533) on SVG elements.

@streamich can you have a look?


robinvw1:
Here's a screenshot where I use `useMeasure` to get the `width` of the SVG `<text>` element and apply it with padding to the `<rect>`.

image

robinvw1:
@xobotyi done

PR #1792, opened 2021-02-04
Default to window.innerWidth instead of 0 in createBreakpoint

### LINKED ISSUE #1493: useBreakpoint always returns the smallest breakpoint while it's calculating the width
`useBreakpoint` always returns the smallest breakpoint for the first time it runs regardless of the actual screen size. This is very inconvenient because if code renders different UI based on this breakpoint then it first renders the UI for a small screen, and then immediately switches to the UI for a larger screen. This makes the UI to flash for a brief moment.

https://codesandbox.io/s/vigilant-hamilton-xhtn4?file=/src/App.js

The expected behavior is that it returns `undefined` until the widow width is calculated. This way, the consumer code can wait until the actual breakpoint is provided.

**A little about versions:**  
- _OS_: All browsers
- _Browser (vendor and version)_: All browsers
- _React_: latest
- _`react-use`_: latest

fnune:
# Description

Fixes #1493 

I tried to write tests for this but I couldn't manage with the setup `react-use` has. I think it's hard to catch in JSDom.

## Type of change

# Checklist
- [x] Add documentation

### DIFF
--- a/src/factory/createBreakpoint.ts
+++ b/src/factory/createBreakpoint.ts
@@ -1,10 +1,10 @@
 import { useEffect, useMemo, useState } from 'react';
-import { off, on } from '../misc/util';
+import { isBrowser, off, on } from '../misc/util';
 
 const createBreakpoint = (
   breakpoints: { [name: string]: number } = { laptopL: 1440, laptop: 1024, tablet: 768 }
 ) => () => {
-  const [screen, setScreen] = useState(0);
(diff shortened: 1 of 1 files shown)


xobotyi:
This won't work server side.
Suppose it should be smth like `const [screen, setScreen] = useState(typeof window !== 'undefined' ? window.innerWidth : 0);`

fnune:
Added an `isBrowser` guard from utils. Thanks @xobotyi!

PR #1819, opened 2021-02-15
Makes createGlobalState API compatible with React's useState API

joaomilho:
# Description

This PR makes the `createGlobalState` function compatible with the React API. This means 2 things:

- It allows using a function in setState, as in `setState(prevState => prevState + 1)`
- It allows using a function in the initial state, as in `createGlobalState(() => 0)`

Also, as a side effect:

- It improves the types when the initial value is passed or set as the generic argument, while relying on the React original type for the return (`SetStateAction`).
- It fixes the current tests that were using an incorrect syntax and extends them to the new use cases and types.
- Extends docs and fixes a minor grammar issue.

## Type of change

# Checklist
   We mobbed this in our team at Spotify.
- [x] Add documentation

### DIFF
--- a/docs/createGlobalState.md
+++ b/docs/createGlobalState.md
@@ -1,6 +1,6 @@
 # `useGlobalState`
 
-A React hook which creates a globally shared state.
+A React hook that creates a globally shared state.
 
 ## Usage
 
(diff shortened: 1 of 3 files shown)


xobotyi:
You better use method provided at `src/misc/hookState.ts` to have a single state resolution method.

joaomilho:
There you go, thanks for the pointer @xobotyi !

PR #1882, opened 2021-03-15
add useConditionalEffect

stutrek:
# Description

This pattern is something we use a lot:
```ts
useEffect(() => {
    if (condition) {
        doSomething(dep);
    }
}, [condition, doSomething, dep])
```

Although there's no better way to do it with hooks, it feels a little dirty.

This PR adds `useConditionalEffect`, which wraps that pattern in a hook.

```ts
useConditionalEffect(boolean, () => {
    doSomething(dep);
}, [doSomething, dep]);
```

There is also `useConditionalLayoutEffect`, which does the same thing with a layout effect.

## Type of change

# Checklist
- [x] Add documentation
- [] Ensure the test suite passes (`yarn test`)


src/useConditionalEffect.ts
```diff
@@ -0,0 +1,14 @@
+import { useEffect, EffectCallback } from 'react';
+
+export const useConditionalEffect = (
+  condition: boolean,
+  effect: EffectCallback,
+  dependencies: any[]
```
oirodolfo, line 6: `DependencyList[]`?

stutrek: Does React export a `DependencyList` type? I thought I checked and saw that it's an array of any. I may be mistaken.

Rycochet: Just plain `DependencyList | undefined` works

Rycochet (commented):
Couple of issues in the code:

The `condition` is a truthy value, so including it in the dependencies array will result in unnecessary re-renders.
Any change in any dependency will result in a re-render, even if `condition` is false.
The array is being built on every render resulting in unnecessary GC thrashing and work.
An unnecessary function is being created on every render.

Of note - arguments are `let` and not `const`, which can be made use of! :-)

stutrek:
There ya go, all concerns are addressed.

Rycochet:
Looks beautiful - final thing: add some tests for multiple dependencies - got a feeling that there might be complaints when the list length changes (`new Array(length)` can be used, and maybe cache it as a `useRef`?)

oirodolfo:
Lovely <3 I'm looking forward to using it as soon as it gets merged! ;)

src/useConditionalEffect.ts
```diff
@@ -0,0 +1,12 @@
...
+
+const FALSE_DEP_ARRAY = [{}];
+const NOOP = () => {};
+
+export const useConditionalEffect = (
+  condition: boolean,
+  effect: EffectCallback,
+  dependencies: DependencyList | undefined
+) => {
+  useEffect(condition ? effect : NOOP, condition ? dependencies : FALSE_DEP_ARRAY);
```
Rycochet, line 11: Pretty sure React will complain when changing the length of the dependencies, so might be worth doing something like this instead? Could also do it by globally caching the fake deps per array length, though this way does handle not having any, while caching would need specific handling

```suggestion
  const fakeDeps = useRef(() => dependencies && new Array(dependencies.length));

  useEffect(condition ? effect : NOOP, condition ? dependencies : fakeDeps.current);
```

stutrek: Good catch, fixed.

.editorconfig
```diff
@@ -11,6 +11,7 @@ trim_trailing_whitespace = true
 max_line_length = 100
 
 [*.{ts, tsx}]
+quote_type = single
```
stutrek, line 14: my VSCode doesn't format correctly without this.

PR #2010, opened 2021-05-18
fix(useTitle): fix the strict mode issue and optimize the document.title access times

realdennis:
# Description

1. Don't think we should default the restore as `false` (more, we don't even need this), since it's more like a effect, if we don't do cleanup, why we use this hook?  If user pass the `restoreOnUnmount` as true, user can just do assign value to `document.title`, it's no any different, but with performance decrease [see below 2.]
2. **Currently in strict mode the `useTitle` does not restore to original title when I pass the `restoreOnUnmount=true`** 
See the demo https://codesandbox.io/s/usetitle-has-bug-hq4ku?file=/src/App.js:131-147
3. There's a performance issue, seems we update the `document.title` in each render.

For 3. I know it'll break the current usage, but I thought we should do before docs update. @streamich 

# What I do

1. **Turn the defaultOption `restoreOnUnmount` to be true**, this should be a side-effect, and I thought we should restore in cleanup as default
2. The current version we update the `document.title` just when the component re-render, it's just redundancy, let's access `document.title` only when `title` update
3. Separate the `title` effect and the restore clean-up to 2 effects, we restore it when unmounted, so the that effect use `[]` as comparedArray

## Type of change

# Checklist
- [ ] Add documentation

### DIFF
--- a/src/useTitle.ts
+++ b/src/useTitle.ts
@@ -5,21 +5,23 @@ export interface UseTitleOptions {
 }
 
 const DEFAULT_USE_TITLE_OPTIONS: UseTitleOptions = {
-  restoreOnUnmount: false,
+  restoreOnUnmount: true,
 };
 
 function useTitle(title: string, options: UseTitleOptions = DEFAULT_USE_TITLE_OPTIONS) {
-  const prevTitleRef = useRef(document.title);
-  document.title = title;
+  const orignalTitleRef = useRef(document.title);
   useEffect(() => {
     if (options && options.restoreOnUnmount) {
       return () => {
-        document.title = prevTitleRef.current;
+        document.title = orignalTitleRef.current;
       };
-    } else {
-      return;
     }
+    return;
   }, []);
+
+  useEffect(() => {
+    document.title = title;
+  }, [title]);
 }
 
-export default typeof document !== 'undefined' ? useTitle : (_title: string) => {};
+export default typeof document !== "undefined" ? useTitle : (_title: string) => {};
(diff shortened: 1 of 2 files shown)


src/useTitle.ts
```diff
@@ -5,21 +5,23 @@ export interface UseTitleOptions {
...
       };
-    } else {
-      return;
     }
+    return;
   }, []);
+
+  useEffect(() => {
+    document.title = title;
+  }, [title]);
```
realdennis, line 24: prevent the redundant assigning in each rendering

src/useTitle.ts
```diff
@@ -5,21 +5,23 @@ export interface UseTitleOptions {
...
 
 const DEFAULT_USE_TITLE_OPTIONS: UseTitleOptions = {
-  restoreOnUnmount: false,
+  restoreOnUnmount: true,
 };
 
 function useTitle(title: string, options: UseTitleOptions = DEFAULT_USE_TITLE_OPTIONS) {
-  const prevTitleRef = useRef(document.title);
-  document.title = title;
+  const orignalTitleRef = useRef(document.title);
```
realdennis, line 12: use the more clear variable name

src/useTitle.ts
```diff
@@ -5,21 +5,23 @@ export interface UseTitleOptions {
 }
 
 const DEFAULT_USE_TITLE_OPTIONS: UseTitleOptions = {
-  restoreOnUnmount: false,
+  restoreOnUnmount: true,
```
realdennis, line 8: it should be true as default, more we don't need this option, if user do not want restore, they could just do assign to `document.title` lol

wait for discussing with owner, and we can preserve(stay in false) or remove this option in later commit.

PR #2033, opened 2021-06-09
fix(useThrottle): Support function values

defunctzombie:
If the caller passes a function as the value, this invokes the function form of
useState and setState which does not set the state to the value but instead invokes
the function and sets the state to the return value of the function. This causes
useThrottle to behave incorrectly when passing a value which is a function. Rather
than returning the function (as the type signature indicates), useThrottle returns
the return value of the function.

This changes fixes useThrottle to support function values as the value to throttle
to align with the type signature.

## Type of change

Depending on how you think existing functionality should behave. The type signature indicates that the value passed to useThrottle is the value that is throttled - but existing callers which previously passed a function and rely on the behavior of calling the function to return the value would no longer be getting said behavior.

# Checklist
- [ ] Add documentation
- [ ] Make sure code lints (`yarn lint`). Fix it with `yarn lint:fix` in case of failure. (yarn lint doesn't run)

### DIFF
--- a/src/useThrottle.ts
+++ b/src/useThrottle.ts
@@ -2,18 +2,18 @@ import { useEffect, useRef, useState } from 'react';
 import useUnmount from './useUnmount';
 
 const useThrottle = <T>(value: T, ms: number = 200) => {
-  const [state, setState] = useState<T>(value);
+  const [state, setState] = useState<T>(() => value);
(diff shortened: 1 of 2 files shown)


defunctzombie:
If the intent is to support functions just like setState supports functions then the alternative change is to update the type annotation.

PR #2034, opened 2021-06-09
fix(useThrottle): add type annotation for values which are functions

defunctzombie:
If the value is a function, it is invoked and the return value of the function
becomes the throttled state. This happens because value is passed directly to
setState which has different behavior for function values.

This change updates the type annotation to correctly reflect the behavior
so users who pass function values get the correct return type (T) from
useThrottle.

## Type of change

# Checklist
- [ ] Add documentation
- [ ] Make sure code lints (`yarn lint`). Fix it with `yarn lint:fix` in case of failure. (yarn lint doesn't run)

### DIFF
--- a/src/useThrottle.ts
+++ b/src/useThrottle.ts
@@ -1,7 +1,7 @@
 import { useEffect, useRef, useState } from 'react';
 import useUnmount from './useUnmount';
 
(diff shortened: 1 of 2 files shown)


defunctzombie:
This is an alternative to https://github.com/streamich/react-use/pull/2033

PR #2036, opened 2021-06-10
fix(createGlobalState): do not mix global states when juggling them i…

qwelias:
…n one component

# Description

In previous configuration states of two global states could get mixed together via `globalState` variable because `useState` remembers the initial `store.state`

## Type of change

# Checklist
- [ ] Add documentation

Here's repro and fix side-to-side
https://codesandbox.io/s/react-playground-forked-013co?file=/index.js


seloner:
How the states can be mixed ? I dont get it

qwelias:
@seloner try to run the test case without the fix :)

seloner:
Also in the current implementation there is no clean up when the hook is unmounted
There is an open pull request from that 
https://github.com/streamich/react-use/pull/1078/files

qwelias:
That's a separate issue I think, and in my usecase with the object map I handle it outside of hook

PR #2046, opened 2021-06-22
Fix: fix useLocalStorage func setState to have correct state

nowells:
# Description

Right now the `useLocalStorage` hook does not properly add `state` as a `useCallback` dependency, leading to bugs where you pass a `function` as the argument to `setState`

## Type of change

# Checklist
- [x] Add documentation

Your `eslint` fails due to configuration changes in `prettier@8` which I am guessing happened a while ago but is not tested against.


BhavyaCodes:
Spent about an hour trying to setState using callback function, ended up creating my own hook for now
```
import { useEffect, useState } from 'react';

export default function useLocalstorage<T>(
  key: string,
  initialVal: T,
): [T, React.Dispatch<React.SetStateAction<T>>] {
  const [state, setState] = useState<T>(initialVal);

  useEffect(() => {
    const localStorageValue = window.localStorage.getItem(key);
    if (localStorageValue !== null) {
      setState(JSON.parse(localStorageValue));
    }
  }, [key]);

  useEffect(() => {
    window.localStorage.setItem(key, JSON.stringify(state));
  }, [state, key]);

  return [state, setState];
}
```

ribx:
This is not enough in every case. If something like set(s => s + 1) is called multiple times, before a rerender, it would produce a wrong result, as the old state is used twice.

src/useLocalStorage.ts
```diff
@@ -79,7 +79,7 @@ const useLocalStorage = <T>(
         // localStorage can throw. Also JSON.stringify can throw.
       }
     },
-    [key, setState]
+    [key, setState, state]
```
reckter, line 82: I think we need to wrap lines 62 to 76 in a `setState` call, to fix the case of @ribx 
something like this: 

```ts
const set: Dispatch<SetStateAction<T | undefined>> = useCallback(
    (valOrFunc) => {
      try {
        setState(oldValue => {
          const newState =
            typeof valOrFunc === 'function' ? (valOrFunc as Function)(oldState) : valOrFunc;
          if (typeof newState === 'undefined') return;
          let value: string;
          if (options)
            if (options.raw)
              if (typeof newState === 'string') value = newState;
              else value = JSON.stringify(newState);
            else if (options.serializer) value = options.serializer(newState);
            else value = JSON.stringify(newState);
          else value = JSON.stringify(newState);
          localStorage.setItem(key, value);
          return deserializer(value)
        }
      } catch {
        // If user is in private mode or has storage restriction
        // localStorage can throw. Also JSON.stringify can throw.
      }
    },
    [key, setState]
  );
  ```

PR #2080, opened 2021-08-31
chore: Fix linting in CI

kachkaev:
# Description

## Type of change

- [ ] ~~Bug fix _(non-breaking change which fixes an issue)_~~
- [ ] ~~New feature _(non-breaking change which adds functionality)_~~
- [ ] ~~**Breaking change** _(fix or feature that would cause existing functionality to not work as before)_~~
- [x] Chore

# Checklist
- [x] Read the [Contributing Guide](https://github.com/streamich/react-use/blob/master/CONTRIBUTING.md)
- [x] Perform a code self-review
- [x] Comment the code, particularly in hard-to-understand areas
- [ ] ~~Add documentation~~
- [ ] ~~Add hook's story at Storybook~~
- [ ] ~~Cover changes with tests~~
- [ ] ~~Ensure the test suite passes (`yarn test`)~~
- [ ] ~~Provide 100% tests coverage~~
- [x] Make sure code lints (`yarn lint`). Fix it with `yarn lint:fix` in case of failure.
- [ ] ~~Make sure types are fine (`yarn lint:types`).~~

PR #2082, opened 2021-08-31
chore: Reconfigure GitHub workflows

kachkaev:
# Description

This PR addresses https://github.com/streamich/react-use/pull/2080#pullrequestreview-743239467.

The yaml is inspired by [gicentre/prettier-plugin-elm](https://github.com/gicentre/prettier-plugin-elm) → [.github/workflows/check-codebase.yml](https://github.com/gicentre/prettier-plugin-elm/blob/c33bf731581f69e19e71a419fe9fed5c9d24e60f/.github/workflows/check-codebase.yml).

## Type of change

- [ ] ~~Bug fix _(non-breaking change which fixes an issue)_~~
- [ ] ~~xNew feature _(non-breaking change which adds functionality)_~~
- [ ] ~~**Breaking change** _(fix or feature that would cause existing functionality to not work as before)_~~
- [x] Chore (CI updates)

# Checklist
- [ ] ~~Add documentation~~
- [ ] ~~Add hook's story at Storybook~~
- [ ] ~~Cover changes with tests~~
- [ ] ~~Provide 100% tests coverage~~


.github/workflows/check-codebase.yml
```diff
@@ -0,0 +1,95 @@
...
+      - name: Check out repository
+        uses: actions/checkout@v2
+
+      - name: Setup node
+        uses: actions/setup-node@v1
+        with:
+          node-version: 14
+
+      - name: Install dependencies
+        run: yarn install --frozen-lockfile
```
kachkaev, line 25: Using `--frozen-lockfile`  instead of `--pure-lockfile` according to https://classic.yarnpkg.com/en/docs/cli/install/

> If you need reproducible dependencies, which is usually the case with the continuous integration systems, you should pass `--frozen-lockfile` flag.

.github/workflows/check-codebase.yml
```diff
@@ -0,0 +1,95 @@
...
+
+      - name: Lint with ESLint
+        run: yarn lint
+
+      - name: Lint with TSC
+        if: ${{ always() }}
```
kachkaev, line 67: If ESLint fails, we still get feedback from TSC.
Ref: https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#job-status-check-functions

streamich:
Maybe, let's drop Windows? It takes too long:

image

streamich (approved):
Looks great, thank you!

kachkaev:
Yeah `yarn install` is not the fastest on Windows. I removed it from the MVP for now, it it might be worth re-adding this OS to the checks later (at least Node LTS). The install might get faster once we’ve upgraded all deps. I suspect that we’re dealing with some binary package that is built on the fly 🤔

Happy to bring Windows back now if you’ve changed your mind 🙂

streamich:
> Happy to bring Windows back now if you’ve changed your mind 🙂

Let's remove it for now, we can always add it later if there is need.

kachkaev:
👍 Done in 2cc19836e02e9c4952f6541cc4cff9c1354ea8b4

streamich:
> 👍 Done in 2cc19836e02e9c4952f6541cc4cff9c1354ea8b4

Great, you should be able to merge it.

kachkaev:
Thanks! Happy for you to merge the PR – I’m not sure if there is a preferred policy:

Screenshot 2021-09-01 at 00 03 14

I usually have this in the repo settings:
Screenshot 2021-09-01 at 00 04 30

kachkaev:
@streamich which merge type would you recommend?

streamich:
> @streamich which merge type would you recommend?

I way is fine, but the final merge should contain the right [semantic commits](https://www.conventionalcommits.org/en/v1.0.0-beta.4/) so that `semantic-release` will automatically release the right version.

If somebody uses semantic Angular-style commits, I would merge using "Merge commit", if not, I would squash and add the right commit type (`feat:`, 'fix:', etc..).

streamich:
@kachkaev Thanks, good improvement!

PR #2140, opened 2021-09-24
useCopyToClipboard: add mimeType param, default it to text/plain

sudhanshu16:
# Description
When you use the hook to copy a text, it copies as `text/html` (since it is default for `copy-to-clipboard` package). This results in weird rendering in rich text editors, example:
image

The value supplied to the method was of type string.

Since almost everyone will be copying plain text using this hook, it makes sense to use `text/plain` as the default value for mimeType.
We can go the other way as well, and set the default to `text/html` but most people will have to supply `text/plain` as a param.
I am actually surprised no one noticed/asked for this before. I assume most people will be using it to copy things like links, emails or plain text. 🤔 

Feel free to correct me if I am wrong with the above assumption.
Looking forward to a review.
Thanks

## Type of change

# Checklist
- [x] Add documentation

### DIFF
--- a/docs/useCopyToClipboard.md
+++ b/docs/useCopyToClipboard.md
@@ -24,9 +24,10 @@ const Demo = () => {
 ## Reference
 
 ```js
-const [{value, error, noUserInteraction}, copyToClipboard] = useCopyToClipboard();
+const [{value, error, noUserInteraction}, copyToClipboard] = useCopyToClipboard(mimeType);
 ```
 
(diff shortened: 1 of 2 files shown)


KushalShah2619:
Hi @streamich, can we please add this mimeType option in `useCopyToClipboard` hook, we can keep `text/html` as default format for copy so there won't be any backward compatibly issue. but without `text/plain` format option, it is creating issue in rich text editor.
Without this option currently, I have to use `copy-to-clipboard` library for only this use case.

lakbychance:
@streamich Having a configurable `mimetype` will help. Facing the same rich-editor issue as the user above. Unless you think it will break backwards compatibility, I think `text/plain` should be safe as default. The `navigator` also defaults to that afaik.

Issue #2165, opened 2021-10-07
useSlider: event callbacks are not updated


erykpiast:
**What is the current behavior?**

With the main effect that attaches all events to the referenced element watching only the single option `vertical` and not the callbacks ([see current code](https://github.com/streamich/react-use/blob/master/src/useSlider.ts#L136)), there is a risk of trapping some outdated state value and calling an invalid function. Consider this example.

https://codesandbox.io/s/react-use-slider-callbacks-not-updated-bug-erfz1

When the `multiplier` state gets updated, the `onScrub` callback isn't internally updated and `setPosition` is always called with the new value multiplied by `50`.

**What is the expected behavior?**
All callbacks are updated accordingly if they change with a subsequent render of the hook.

**A little about versions:**  
- _OS_: MacOS X
- _Browser (vendor and version)_: Firefox 93
- _React_: 17.0.2
- _`react-use`_: 17.3.1 
- _Did this worked in the previous package version?_ No

PR #2166, opened 2021-10-07
fix(useSlider): fix dependency array for the effect

erykpiast:
# Description

Closes https://github.com/streamich/react-use/issues/2165. See the issue for more details.

Putting all used values inside the dependency array fixes that issue however it might introduce some performance regression, depending on how often callbacks actually change.

For consistency, `isMounted` and `setState` are added to the array as well. They shouldn't change in theory so they don't cause any issue but even if case they do change, proper reference will be taken thanks to adding them to this array.

## Type of change

# Checklist
- [x] Add documentation

### DIFF
--- a/src/useSlider.ts
+++ b/src/useSlider.ts
@@ -133,7 +133,7 @@ const useSlider = (ref: RefObject<HTMLElement>, options: Partial<Options> = {}):
     } else {
       return undefined;
     }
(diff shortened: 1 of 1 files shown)


vincaslt:
it's been 3 years, maybe time to merge?

PR #2215, opened 2021-11-06
perf: use performance.now and RAF callback argument (time) for useRaf.

SCG82:
# Description

* `Date.now()` is not guaranteed to be monotonic, `performance.now()` is.
* Avoid calling `Date.now()` on every RAF by utilizing the provided RAF callback argument: a DOMHighResTimeStamp similar to the one returned by `performance.now()`, indicating the point in time when `requestAnimationFrame()` starts to execute callback functions. https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame

## Type of change

# Checklist
- [x] Add documentation


src/useRaf.ts
```diff
@@ -22,7 +21,7 @@ const useRaf = (ms: number = 1e12, delay: number = 0): number => {
         cancelAnimationFrame(raf);
         set(1);
       }, ms);
-      start = Date.now();
+      start = performance.now();
```
streamich, line 24: What do you think about keeping by default the original behavior, but allowing the user to overwrite it?

```suggestion
      start = now();
```

where

```ts
const useRaf = (ms: number = 1e12, delay: number = 0, now = Date.now): number => {
```

SCG82: To allow configuration of the starting time? Isn't that what the delay is for?

Issue #1885, opened 2021-03-16
When I refresh the page useMedia will break my UI (Remove css)


jemink:
**I implemented useMedia() from `react-use` so when I check desktop to mobile then it is working fine but when I refresh on mobile then it will remove my css**

**
import React, { FC } from 'react'
import { useMedia } from 'react-use'

const Index: FC = () => {
  const isMobile = useMedia('(max-width: 768px)', false);
  
  return (
    <div>
      {isMobile ? (<div className={styles.mobileView}> Mobile View </div>) : (<div className={styles.desktopView}> Desktop View </div>)}
    </div>
    )
}

.mobile-view {
 background-color: aquamarine !important;
}

.desktop-view {
  background-color: red;
}
**

**Bug issue video:- https://www.awesomescreenshot.com/video/3078205?key=fd357595486a738f6fff5fea636a7597**

**Version**  
- "less": "^3.13.0" 
-  "react": "17.0.1" 
- "react-use": "^17.1.1" 
- "next": "10.0.7"

jemink:
In above video you can see that when I refresh on the mobile view it will apply style of desktop view that is `background-color: red;`

nghiepit:
@jemink 

Maybe it's because of React 16's new SSR strategy? https://github.com/facebook/react/issues/10591.
And in PR [useMedia - initialize state with call to media query](https://github.com/streamich/react-use/pull/454) make initialize state difference on server and client.

garryburch:
Same issue here

arifintajul4:
same to me

mvshmakov:
Shouldn't this issue be closed with https://github.com/streamich/react-use/pull/2216?

PR #2216, opened 2021-11-08
fix: useMedia SSR hydration bug with defaultState

teaguestockwell:
Fix server side render hydration mismatch when a default value is provided to useMedia

# Description

Before, in pr #464 when server side rendering a component that used useMedia, the server's initial state would initialize to false, but then when React hydrated this on the client, it would initialize to the result of the media query. This caused bug #1885 because When React hydrates the server render, it may not match the client's state. The [React docs](https://reactjs.org/docs/react-dom.html#hydrate) have a little more info on why this can lead to perf issues as well.

## Type of change

- [X] Bug fix _(non-breaking change which fixes an issue)_

# Checklist
- [X] Add documentation

### DIFF
--- a/docs/useMedia.md
+++ b/docs/useMedia.md
@@ -25,3 +25,5 @@ useMedia(query: string, defaultState: boolean = false): boolean;
 ```
 
 The `defaultState` parameter is only used as a fallback for server side rendering.
+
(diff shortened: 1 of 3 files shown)


streamich (approved):
Thank you!

PR #2266, opened 2021-12-07
Pr/use local storage fix

cheemcheem:
# Description

The lack of these dependencies in the set function in the useCallback hook on line 59 (especially the lack of the 'state' dependency) was causing incorrect behaviour.

E.g. when the set method is called with a function, it would use an out of date state and would therefore apply the set function to the old state, causing an unexpected result.

Note, the setState dependency was redundant and was removed, see [react docs](https://reactjs.org/docs/hooks-reference.html#usestate):
 _"React guarantees that setState function identity is stable and won’t change on re-renders. This is why it’s safe to omit from the useEffect or useCallback dependency list."_

Also added test to verify that the change worked. The test makes sure that calling setState with a function twice in a row will preserve both the first and second state changes. 

## Type of change

# Checklist
- [ ] Add documentation **(reason: fixing a bug, intended functionality was already documented)**
- [ ] Add hook's story at Storybook **(reason: fixing a bug, I assume this was already done for this hook)**
- [ ] Provide 100% tests coverage **(reason: coverage unchanged and was already below 100%)**

### DIFF
--- a/src/useLocalStorage.ts
+++ b/src/useLocalStorage.ts
@@ -79,7 +79,7 @@ const useLocalStorage = <T>(
         // localStorage can throw. Also JSON.stringify can throw.
       }
     },
-    [key, setState]
+    [deserializer, key, options, state]
   );
 
   // eslint-disable-next-line react-hooks/rules-of-hooks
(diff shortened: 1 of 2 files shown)


ribx:
This is not enough in every case. If something like set(s => s + 1) is called multiple times, before a rerender, it would produce a wrong result, as the old state is used twice.

Issue #2273, opened 2021-12-17
Unexpected behavior of `useLocalStorage` with objects 


41y08h:
[This is worth more than all that could have been written here](https://codesandbox.io/s/quirky-wilbur-bcswp?file=/src/App.js)

brickspert:
The [useLocalStorageState](https://ahooks.js.org/zh-CN/hooks/use-local-storage-state) of ahooks work fine. It seems that react-use is no longer maintained.

https://codesandbox.io/s/bold-mountain-ekwbp?file=/src/App.js

PR #2277, opened 2022-01-08
fix(useLocalStorage): fix issue #2273

Rabi-Wang:
useCallback missing dependent state

# Description

## Type of change

# Checklist
- [ ] Add documentation


ribx:
This is not enough in every case. If something like `set(s => s + 1)` is called multiple times, before a rerender, it would produce a wrong result, as the old state is used twice.

PR #2286, opened 2022-01-19
rewrite useMap hook

MichalTarasiuk:


src/useMap.ts
```diff
@@ -1,47 +1,61 @@
...
-}
+const createMap = <TValue>(value: TValue): Map<keyof TValue, any> =>
+  Object.keys(value).reduce((acc, key) => {
+    acc.set(key, value[key]);
 
-export interface Actions<T extends object> extends StableActions<T> {
-  get: <K extends keyof T>(key: K) => T[K];
-}
+    return acc;
+  }, new Map());
```
wangxingkai, line 8: we could do this instead: `new Map(Object.entries(value))`

wangxingkai: oh, but the `Object.entries` will force a string type key

Issue #1245, opened 2020-05-28
useLocalStorage doesn't fire when localStorage is changed from another tab


vadzim:
I believe `useLocalStorage` should fire and update its value when local storage is changed from another browser tab, isn't it?

zhengbigbig:
Component which used useLocalStroage should be forced to mount again.

zhengbigbig:
If you read and write it frequently, suggest you encapsulate it yourself, It could be costly.

sungwoo91:
I think useLocalStorage just modify the value in the window.localStorage. So you should register storage eventListener to update another tab page.

Like below

```tsx
const Demo = () => {
  const [value, setValue] = useLocalStorage('hello-key', 'foo', {raw: true});
  const [removableValue, setRemovableValue, remove] = useLocalStorage('removeable-key');

  const callback = React.useCallback((e) => {
    setValue(e.newValue);
  }, [setValue]);

  useEvent('storage', callback);
  
  return (
    <div>
      <div>Value: {value}</div>
      <button onClick={() => setValue('bar')}>bar</button>
      <button onClick={() => setValue('baz')}>baz</button>
      <br />
      <br />
      <div>Removable Value: {removableValue}</div>
      <button onClick={() => setRemovableValue('foo')}>foo</button>
      <button onClick={() => setRemovableValue('bar')}>bar</button>
      <button onClick={() => remove()}>Remove</button>
    </div>
  );
};
```

Svish:
I tried registering for storage events in a local `useLocalStorage` variant of mine, and... it's _very_ easy to end up in infinite update-loops. If it's going to be a feature in this library, it should be _very_ well tested.

xobotyi:
If anyone interested it is implemented in upconig hook of `@react-hookz/web` https://github.com/react-hookz/web/pull/43

innrvoice:
Draft PR for possible solution #2307

innrvoice:
@streamich fyi

innrvoice:
Can anyone ping one of the maintainers or contributors. My pr stuck in workflow because I am "First-time contributor". 
@vadzim @Svish @sungwoo91 

Thanks!

Svish:
@innrvoice It's not stuck because you're a "First-time contributor", it's stuck because this project is kind of dead. If you look at the history, there haven't been any proper PR's merged for a long time.

innrvoice:
@Svish thank you for the answer. Its a pity that this project is no longer needed.

PR #2307, opened 2022-03-23
feat: local storage updates from other "contexts"  (draft)

innrvoice:
# Description

If local storage value changes in some other "context" (e.g. other component...), hook value will update on next render.

## Type of change

- [x ] Bug fix _(non-breaking change which fixes an issue)_

# Checklist
- [x ] Read the [Contributing Guide](https://github.com/streamich/react-use/blob/master/CONTRIBUTING.md)
- [?] Add documentation
- [?] Add hook's story at Storybook
- [?] Provide 100% tests coverage

### DIFF
--- a/src/useLocalStorage.ts
+++ b/src/useLocalStorage.ts
@@ -23,6 +23,8 @@ const useLocalStorage = <T>(
     throw new Error('useLocalStorage key may not be falsy');
   }
 
+  const localStorageValue = localStorage.getItem(key); // current local storage value on every hook render
+
   const deserializer = options
     ? options.raw
       ? (value) => value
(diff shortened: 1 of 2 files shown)


innrvoice:
@streamich fyi

colgin:
@innrvoice This pr didn't fix the issue mentioned in https://github.com/streamich/react-use/issues/1245, In this implementation, we get the latest storage value in every re-render, but storage value changing doesn't make every component re-render.

```js
<Component1 /> {/* use the storage with key "foo", and change the value */}
<Component2 /> {/* use the storage with key "foo", it won't re-render when Component1 re-render, so the localStorageValue  is not the latest*/}
```

PR #2336, opened 2022-05-20
fix:react-hooks/rules-of-hooks in useSize hook

v-kryuchkov:
# Description

## Type of change

# Checklist
- [x] Add documentation

### DIFF
--- a/src/useSize.tsx
+++ b/src/useSize.tsx
@@ -16,22 +16,8 @@ const useSize = (
   element: Element,
   { width = Infinity, height = Infinity }: Partial<State> = {}
 ): [React.ReactElement<any>, State] => {
-  if (!isBrowser) {
-    return [
-      typeof element === 'function' ? element({ width, height }) : element,
-      { width, height },
-    ];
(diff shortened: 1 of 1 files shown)


src/useSize.tsx
```diff
@@ -16,22 +16,8 @@ const useSize = (
   element: Element,
   { width = Infinity, height = Infinity }: Partial<State> = {}
 ): [React.ReactElement<any>, State] => {
-  if (!isBrowser) {
-    return [
-      typeof element === 'function' ? element({ width, height }) : element,
-      { width, height },
-    ];
-  }
```
streamich, line 24: IMO, the browser check should stay at the top. Event better would be if we didn't have the browser check in this hook, but instead had two separate hooks

```ts
const useSize = isBrowser ? useSizeBrowser : useSizeServer;
export default useSize;
```

v-kryuchkov: Good idea. I splitted useSize to useSizeBrowser and useSizeServer

PR #2348, opened 2022-06-08
useEffectOnce update/fix

frederick-pi:
Making sure useEffectOnce actually runs only once. Particularly preventing React 18 from rendering twice in development mode.

### DIFF
--- a/src/useEffectOnce.ts
+++ b/src/useEffectOnce.ts
@@ -1,7 +1,17 @@
-import { EffectCallback, useEffect } from 'react';
+import { EffectCallback, useEffect, useRef } from 'react';
 
 const useEffectOnce = (effect: EffectCallback) => {
-  useEffect(effect, []);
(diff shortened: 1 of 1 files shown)


mrclay:
Take a look at https://dev.to/iamyoki/comment/1pfc9 

```js
function useEffectOnce(effect) {
  const effectFn = useRef(effect)
  const destroyFn = useRef()
  const effectCalled = useRef(false)
  const rendered = useRef(false)
  const [, refresh] = useState(0)

  if (effectCalled.current) {
    rendered.current = true
  }

  useEffect(() => {
    if (!effectCalled.current) {
      destroyFn.current = effectFn.current()
      effectCalled.current = true
    }

    refresh(1)

    return () => {
      if (rendered.current === false) return
      if (destroyFn.current) destroyFn.current()
    }
  }, [])
}
```

This seems to be designed to skip a render, but it's not clear why.

PR #2354, opened 2022-06-20
feat: `useAsync` returns synchronously when given a sync function

simondesancha:
# Description

Firstly, thanks for the library! It has saved me a ton of time, and also introduced me to a bunch of new ideas.

This changes the behaviour of the `useAsync` when given a sync function. 

Previously, this situation would cause an unnecessary render: first `useAsync` will return `loading: true` & `value: undefined`, and then re-render with `loading: false` & `value: "return value of sync function"`

This changes the behaviour, now `useAsync` will immediately return `loading: false` & `value: "return value of sync function"` when given the sync function.

This is useful for situations where is is not obvious whether an operation will be sync or async; e.g. an API request may be cached  & returns synchronously, or it may not be cached & requires `await`-ing.

While this is a behaviour change, I wouldn't expect it to negatively affect current usage of this hook, so I wouldn't call it a breaking change.

## Type of change

# Checklist
- [X] Add documentation

### DIFF
--- a/docs/useAsync.md
+++ b/docs/useAsync.md
@@ -1,7 +1,9 @@
 # `useAsync`
 
 React hook that resolves an `async` function or a function that returns
-a promise;
+a promise. 
+
+If the function returns a value synchronously, `useAsync` will also return the value synchronously.
 
 ## Usage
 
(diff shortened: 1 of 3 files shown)


simondesancha:
hi! are @streamich or any other maintainer's able to take a look at this?

PR #2445, opened 2022-12-01
allow function passed to useObservable

mattkime:
# Description

In response to - https://github.com/elastic/kibana/pull/146638/files#r1035809203

It would be nice to be able to `useObservable(observable$.pipe(...))` without creating a new observable on each render so I'm suggesting supporting `useObservable(() => observable$.pipe(...))`

## Type of change

# Checklist
- [x] Add documentation

### DIFF
--- a/src/useObservable.ts
+++ b/src/useObservable.ts
@@ -7,9 +7,13 @@ export interface Observable<T> {
   };
 }
 
-function useObservable<T>(observable$: Observable<T>): T | undefined;
-function useObservable<T>(observable$: Observable<T>, initialValue: T): T;
-function useObservable<T>(observable$: Observable<T>, initialValue?: T): T | undefined {
+type TypeOrFnReturnsType<T> = T | (() => T);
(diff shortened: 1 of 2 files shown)


streamich:
It seems that if `observable$` will change, it will not re-`.subscribe()` inside the `useObservable`? Which I think should still work.

Alternatively, instead of adding this to `useObservable` you could wrap your call sites in `useMemo`, wdyt?

```ts
useObservable(useMemo(() => observable$.pipe(...)))
```

You could have a local helper:

```ts
const useObservableStatic = lazy => useObservable(useMemo(lazy));
```

and then:

```ts
useObservableStatic(() => observable$.pipe(...))
```

---

One thing that also comes to mind, is that the Rx operator chaining `.pipe(...)` maybe better should be in the controller, not in the React components, so that `.pipe(...)` appear inside React code as little as possible.

mattkime:
>One thing that also comes to mind, is that the Rx operator chaining .pipe(...) maybe better should be in the controller, not in the React components

I agree with this strategy

PR #2458, opened 2022-12-20
fix(useMethods): make returned `methods` object reference stable

### LINKED ISSUE #1286: `useMethods` methods object is not stable and changes when `initialState` is changed
### What is the current behavior?

The `useMethods` hook wraps the behaviour of `useReducer` to return a nicely typed `methods` object. 

https://github.com/streamich/react-use/blob/0fabbcd641e5fc6d16091e7c4c803efe80d7ff56/docs/useMethods.md#L44

However, the `methods` object returned from `useMethods` is not stable and changes whenever the `initialState` argument changes.

This does not match how React treats an `initialState` argument for `useState` and `useReducer`
>The initialState argument is the state used during the initial render. In subsequent renders, it is disregarded. (https://reactjs.org/docs/hooks-reference.html#lazy-initial-state)

This also does not match how React guarantees the `setState` method from `useState` and the `dispatch` method from `useReducer` to both be stable and not to change on re-renders or when `initialState` changes.

### What is the expected behavior?

I would expect that the `methods` object was guaranteed to be stable and not to change on re-renders or when `initialState` changes. (Though I would expect it to change if the `createMethods` argument changed)

This would then match the behaviour of the `setState` method from `useState` and the `dispatch` method from `useReducer`.

### Investigation

Looking at the source of [`useMethods.ts`](https://github.com/streamich/react-use/blob/0fabbcd641e5fc6d16091e7c4c803efe80d7ff56/src/useMethods.ts), I think the current behaviour is unintentional. 

For example, the dispatch behaviour is already stable, since `createMethods` is re-run for each dispatch call:

https://github.com/streamich/react-use/blob/0fabbcd641e5fc6d16091e7c4c803efe80d7ff56/src/useMethods.ts#L19-L24

But the way that the `wrappedMethods` object is created introduces a dependency on `initialState`:

https://github.com/streamich/react-use/blob/0fabbcd641e5fc6d16091e7c4c803efe80d7ff56/src/useMethods.ts#L28-L35

I think the bug was introduced accidentally in https://github.com/streamich/react-use/commit/b394f3d72356d331dbce48acd3686bbb64d331b5, since it wasn't present in the original version of the hook.

### Version Info 
- _OS_: Windows 10
- _Browser_: Chrome 83
- _React_: 16.13.1
- _`react-use`_: 15.2.1, >=13.26.0

danguilherme:
# Description

Make `useMethods` consistent with both `useState` and `useReducer` by not updating the returned `methods` object reference if `initialState` changes after the initial render:

> The `initialState` argument is the state used during the initial render. In subsequent renders, it is disregarded. 
> 
> — [reactjs.org/docs/hooks-reference.html#lazy-initial-state](https://reactjs.org/docs/hooks-reference.html#lazy-initial-state)

Closes #1286 

## Type of change

# Checklist
- [ ] Add documentation

### DIFF
--- a/src/useMethods.ts
+++ b/src/useMethods.ts
@@ -26,14 +26,20 @@ const useMethods = <M, T>(
 
   const [state, dispatch] = useReducer<Reducer<T, Action>>(reducer, initialState);
 
-  const wrappedMethods: WrappedMethods<M> = useMemo(() => {
-    const actionTypes = Object.keys(createMethods(initialState));
+  const wrappedMethods: WrappedMethods<M> = useMemo(
+    () => {
+      const actionTypes = Object.keys(createMethods(initialState));
 
-    return actionTypes.reduce((acc, type) => {
-      acc[type] = (...payload) => dispatch({ type, payload });
-      return acc;
-    }, {} as WrappedMethods<M>);
-  }, [createMethods, initialState]);
+      return actionTypes.reduce((acc, type) => {
+        acc[type] = (...payload) => dispatch({ type, payload });
+        return acc;
+      }, {} as WrappedMethods<M>);
+    },
+    // Do not track `initialState` as it's supposed to be used only once.
+    // See https://github.com/streamich/react-use/issues/1286
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [createMethods]
+  );
 
   return [state, wrappedMethods];
 };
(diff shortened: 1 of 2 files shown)


danguilherme:
@streamich sorry for the tagging, but is there any way this PR can be merged and released?

Issue #2343, opened 2022-06-01
React 18 breaks some usage of `useThrottle`


jpwilliams:
**What is the current behavior?**

React 18 has introduced some [new Strict Mode behaviours](https://reactjs.org/blog/2022/03/29/react-v18.html#new-strict-mode-behaviors) that results in components unmounting and remounting whenever it mounts for the first time, restoring the previous state on the second mount.

For `useThrottle`, this results in the hook believing a timer has already been set and it will wait indefinitely for the now-removed timer to call back. It appears this is caused by the timer being cleared via `clearTimeout` but not unset.

**Steps to reproduce it and if possible a minimal demo of the problem. Your bug will get fixed much faster if we can run your code and it doesn't have extra dependencies other than `react-use`. Paste the link to your [JSFiddle](https://jsfiddle.net) or [CodeSandbox](https://codesandbox.io) example below:**

I'll venture to add a repro repo if needed.

**What is the expected behavior?**

The hook fully clears away the timer on unmount and can remount and start it again.

**A little about versions:**  
- _OS_: Ubuntu
- _Browser (vendor and version)_: Microsoft Edge Version 101.0.1210.53 (Official build) (64-bit)
- _React_: `18`
- _`react-use`_: `17.3.2`
- _Did this worked in the previous package version?_ Not with React 18.

---

That particular fix is an easy one; we change [useThrottle.ts:30](https://github.com/streamich/react-use/blob/master/src/useThrottle.ts#L30):

```ts
// from
timeout.current && clearTimeout(timeout.current);

// to
if (timeout.current) {
  clearTimeout(timeout.current);
  timeout.current = undefined;
}
```

With this, when the state is restored it understands that the timer has been fully cleared away from the initial mount and we're good to go.

In terms of tests, we may be a bit blocked. It looks like [@testing-library/react-hooks](https://github.com/testing-library/react-hooks-testing-library) may be [deprecated](https://github.com/testing-library/react-hooks-testing-library/issues/849) before they add React 18 support. [@testing-library/react@13](https://github.com/testing-library/react-testing-library/releases/tag/v13.0.0) seems to take it's place by adding [their own `renderHook`](https://github.com/testing-library/react-testing-library/pull/991) function, but this also [drops support for React 17](https://github.com/testing-library/react-testing-library/releases/tag/v13.0.0) in the process. 😬

### Related

- testing-library/react-hooks-testing-library#849
- testing-library/react-testing-library#991
[...]

PR #2475, opened 2023-02-08
fix: useThrottle useThrottleFn unmount issue in Next.js

johnnywang1994:
## Description
- Env: next 13.1.1 + Typescript
- React strictMode cause useThrottle, useThrottleFn unmount bug

## Situation:
- As is: click div but nothing print out in console.
- To be: print text in console.
```ts
import { FC, useState } from 'react';
import { useThrottleFn } from 'react-use';

const RefreshPageButton: FC = () => {
  const [value, setVal] = useState(0);

  useThrottleFn(async () => {
    console.log('throttle triggered');
  }, 2000, [value]);

  return (
    <div onClick={() => setVal(value + 1)}>Just Test</div>
  );
};

export default RefreshPageButton;
```
after exploring the reason, I've found that because of the StrictMode, `useThrottleFn` will trigger as `mount`-`unmount`-`mount` in Next.js when first time rendering, which cause the `timeoutCallback` not triggered as expected and let the `timeout` ref value mistakenly remain exist.

## Solution
```ts
import { useEffect, useRef, useState } from 'react';
import useUnmount from './useUnmount';

const useThrottleFn = <T, U extends any[]>(fn: (...args: U) => T, ms: number = 200, args: U) => {
  const [state, setState] = useState<T | null>(null);
  const timeout = useRef<ReturnType<typeof setTimeout>>();
  const nextArgs = useRef<U>();

  useEffect(() => {
    if (!timeout.current) {
      setState(fn(...args));
      const timeoutCallback = () => {
        if (nextArgs.current) {
          setState(fn(...nextArgs.current));
          nextArgs.current = undefined;
          timeout.current = setTimeout(timeoutCallback, ms);
        } else {
          timeout.current = undefined;
        }
      };
      timeout.current = setTimeout(timeoutCallback, ms);
    } else {
      nextArgs.current = args;
    }
  }, args);

  useUnmount(() => {
    timeout.current && clearTimeout(timeout.current);
    // reset timeout ref value to let `timeoutCallback` run as expected in StrictMode
    timeout.current = undefined;
  });

  return state;
};

export default useThrottleFn;
```

## Type of change
- [x] Bug fix useThrottle, useThrottleFn timeout ref value remain exist after first unmount in StrictMode.

# Checklist
- [x] Add documentation

### DIFF
--- a/src/useThrottle.ts
+++ b/src/useThrottle.ts
@@ -28,6 +28,7 @@ const useThrottle = <T>(value: T, ms: number = 200) => {
 
   useUnmount(() => {
     timeout.current && clearTimeout(timeout.current);
+    timeout.current = undefined;
   });
 
   return state;
--- a/src/useThrottleFn.ts
+++ b/src/useThrottleFn.ts
@@ -26,6 +26,7 @@ const useThrottleFn = <T, U extends any[]>(fn: (...args: U) => T, ms: number = 2
 
   useUnmount(() => {
     timeout.current && clearTimeout(timeout.current);
+    timeout.current = undefined;
   });
 
   return state;


royvandewater:
Looks like this fixes #2343 and #2488. Am curious why this hasn't been merged yet.

royriver23:
Agreed. It would be awesome to have this merged 🙏🏼

PR #2485, opened 2023-03-08
feat: add support for a cleanup function in useInterval

RJWadley:
# Description

This change updates the useInterval hook to allow the callback function to optionally return a cleanup function. The cleanup function, if provided, will be called between intervals and on unmount, much like the cleanup function in useEffect.

This change adds more flexibility to the useInterval hook, allowing users to more easily perform additional cleanup or tear-down logic as needed. This makes it easier to work with async code and event listeners.

## Type of change

# Checklist
- [x] Add documentation

### DIFF
--- a/docs/useInterval.md
+++ b/docs/useInterval.md
@@ -1,6 +1,10 @@
 # `useInterval`
 
-A declarative interval hook based on [Dan Abramov's article on overreacted.io](https://overreacted.io/making-setinterval-declarative-with-react-hooks). The interval can be paused by setting the delay to `null`.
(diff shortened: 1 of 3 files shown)


RJWadley:
This change is breaking because the type signature changed from `useInterval(callback: Function, delay?: number | null)` to `useInterval(callback: () => VoidFunction | void, delay?: number | null)`

here's an example that would break:
```tsx
const sampleCallback = () => "Hello!"
useInterval(sampleCallback)
```
This was never technically correct, but the type was relaxed enough to allow it.

Issue #2328, opened 2022-05-09
[Bug report] Remove 


SukkaW:
https://github.com/streamich/react-use/blob/3685b7502a4d1980953deed11c3dee31b220c89d/src/useLatest.ts#L3-L7

Although the current React documentation doesn't mention it, according to [the React beta documentation](https://beta.reactjs.org/apis/useref):

> Do not write or read `ref.current` during rendering.
> React expects that the body of your component behaves like a pure function:
> - If the inputs (props, state, and context) are the same, it should return exactly the same JSX.
> - Calling it in a different order or with different arguments should not affect the results of other calls.
>
> Reading or writing a ref during rendering breaks these expectations.
>
> When you break these rules, your component might still work, but most of the newer features we’re adding to React will rely on these expectations.

Ebuall:
When I was writing something like this, I would wrap it in `useEffect` without thinking. But I don't see how idempotent writing could hurt here.
Reading from ref is a completely different story.

zhouzi:
Wouldn't it be fixed by a `useLayoutEffect`?

```typescript
const useLatest = <T>(value: T): { readonly current: T } => { 
   const ref = useRef(value); 
   useLayoutEffect(() => ref.current = value, [value]);
   return ref; 
 }; 
```

pastelmind:
I use these two hooks frequently in my projects:

```ts
interface LatestRef<T> {
  readonly current: T
}

const INITIAL_VALUE = Symbol('initial value')

function useLatestRef<T>(value: T): LatestRef<T> {
  const ref = useRef(value)
  useEffect(() => {
    ref.current = value
  })
  return ref
}

function useLayoutLatestRef<T>(value: T): LatestRef<T> {
  const ref = useRef(value)
  useLayoutEffect(() => {
    ref.current = value
  })
  return ref
}
```

These hooks update the ref value at different timings, which can be important when the ref is used by another `useEffect()` or `useLayoutEffect()`.

Since people rely on `useLatest()`, perhaps we could deprecate it (with a warning that mentions concurrency and tearing problems) and add these two hooks as separate APIs?

PR #2509, opened 2023-07-14
Add useLatestRef() and useLayoutLatestRef()

pastelmind:
# Description

This PR adds two hooks `useLatestRef()` and `useLayoutLatestRef()` as described in https://github.com/streamich/react-use/issues/2328#issuecomment-1633958158. They are similar to `useLatest()`, but avoid updating the ref directly during render, [which is deemed unsafe by the React team.](https://react.dev/learn/referencing-values-with-refs#best-practices-for-refs).

This PR also marks `useLatest()` as `@deprecated` and recommend `useLatestRef()` as an alternative.

## Type of change

# Checklist
- [x] Add documentation


nmoinvaz:
Nice PR!

nathggns:
Really hope this is merged. useLatest is not safe.

hengkx:
Is there any progress?

wijionejs:
@pastelmind 

Sorry for duplicate message from another thread, but:

What if a child component tried to call a function wrapped using your approach? Seems like it would get stale callback.
Effects run bottom-up (from children to parent), and let's say you update a counter in the parent component from 0 to 1, this would cause the child's useEffect to run first, but since the parent's useEffect hasn't been executed yet, it's closed over the previous value, namely 0.

Am I wrong here?

Issue #2561, opened 2024-05-23
Add useTitle option to restore original on unmount 


StructuralRealist:
**Is your feature request related to a problem? Please describe.**
When implementing `useTitle` you'd usually want to show a default when a page has not implemented the hook yet. 

**Describe the solution you'd like**
An option `restoreOriginalOnUnmount` which will revert to the first title value.

**Describe alternatives you've considered**
`restoreOnUnmount` does a similar thing, but it reverts to the previous title, not the original one. This is actually not so useful for pages: if you go from `About` to `Profile` to `Home` and the latter doesn't set a title, it will restore to `About`.

PR #2563, opened 2024-06-15
feat: adding restoreOriginalOnUnmount to useTitle 

gavinbm:
# Description

New option added to the UseTitleOptions interface that will restore the first title passed into useTitle, as requested in issue #2561. I used useMemo to memoize it, I also added one test case to make sure it would pass. One issue that I need help with is determining what to do if both restoreOnUnmount and restoreOriginalOnUnmount are both true at the same time. I'm not sure how that should be handled.

## Type of change

- [ x ] New feature _(non-breaking change which adds functionality)_

# Checklist
- [ x ] Read the [Contributing Guide](https://github.com/streamich/react-use/blob/master/CONTRIBUTING.md)
- [ x ] Perform a code self-review
- [ x ] Comment the code, particularly in hard-to-understand areas
- [ ] Add documentation
- [ x ] Cover changes with tests
- [ x ] Ensure the test suite passes (`yarn test`)
- [ x ] Make sure code lints (`yarn lint`). Fix it with `yarn lint:fix` in case of failure.
- [ x ] Make sure types are fine (`yarn lint:types`).

### DIFF
--- a/src/useTitle.ts
+++ b/src/useTitle.ts
@@ -2,26 +2,37 @@ import { useEffect, useRef } from 'react';
 
 export interface UseTitleOptions {
   restoreOnUnmount?: boolean;
+  restoreOriginalOnUnmount?: boolean;
 }
 
 const DEFAULT_USE_TITLE_OPTIONS: UseTitleOptions = {
   restoreOnUnmount: false,
+  restoreOriginalOnUnmount: false,
 };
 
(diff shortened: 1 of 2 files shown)


StructuralRealist:
Hi @gavinbm,

I agree having both options makes little sense. You either want to restore to the previous, the default or not at all. What about an enum instead of a boolean?

`restore`: `'previous' | 'default' | 'off'` (or `null`/`undefined`)

PR #2572, opened 2024-07-20
fix: fixed useStateWithHistory issue with history state being a function after running history.back

### LINKED ISSUE #2568: useStateWithHistory: initialState of function type are not handled correctly
With the hook `useStateWithHistory`, when using a function for the `initialState`, the function is correctly called and its output returned as the first `state` value

But when using the `back` handler, the `state` returned is the function and not its output.

https://codesandbox.io/p/sandbox/withered-sunset-gwjw8r

In case the `initialState` is a function, we should return its output, always, and not even store the function, as per the [React Dev documentation](https://react.dev/reference/react/useState#parameters) stating this function should be triggered only once:
> If you pass a function as initialState, it will be treated as an initializer function. It should be pure, should take no arguments, and should return a value of any type. React will call your initializer function when initializing the component, and store its return value as the initial state. [See an example below.](https://react.dev/reference/react/useState#avoiding-recreating-the-initial-state)

**A little about versions:**  
- _OS_: N/A
- _Browser (vendor and version)_: N/A
- _React_: N/A
- _`react-use`_: latest
- _Did this worked in the previous package version?_: No

akdlsz21:
# Description

useStateWithHistory had an issue when the initialState is a function.
The back function breaks due to the history's first value remaining as a function. 

I fixed this so the resolved value to be stored in history. 

fixes #2568

## Type of change

# Checklist
- [ ] Add documentation

### DIFF
--- a/src/useStateWithHistory.ts
+++ b/src/useStateWithHistory.ts
@@ -30,7 +30,8 @@ export function useStateWithHistory<S, I extends S>(
   }
 
   const isFirstMount = useFirstMountState();
-  const [state, innerSetState] = useState<S>(initialState as S);
+  const [state, innerSetState] = useState<S>(resolveHookState(initialState) as S);
+
   const history = useRef<S[]>((initialHistory ?? []) as S[]);
   const historyPosition = useRef(0);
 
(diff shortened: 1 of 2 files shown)


akdlsz21:
Hi @streamich — friendly ping on this bugfix. 

It corrects `useStateWithHistory` when initialState is a function (prevents back() returning the initializer). 
Closes #2568.

Happy to adjust anything you’d like. Thank you for your time and for maintaining react-use!

PR #2596, opened 2024-09-27
build: support ES modules

### LINKED ISSUE #2353: react-use cannot be imported in Node.js

I can't use ESM syntax to import `react-use` in Node.js. I found this issue when I try to run some unit tests on Node.js. 

The reason for this issue is that the following package.json is not supported by Node.js. There is a detailed explanation https://github.com/sheremet-va/dual-packaging

```
{
  "name": "react-use",
  "main": "lib/index.js",
  "module": "esm/index.js",
  ...
}
```

There are some simple methods to fix this issue. More explanation in the link above.

1. Put a `package.json` under `esm/`, with the content `{"type": "module"}`;
2. Or, rename all `.js` file under `esm/` to `.mjs`;
3. Or. rename all `.js` file under `lib/` to `.cjs`.

```
bash-3.2$ pnpm add react-use
dependencies:
+ react-use 17.4.0
bash-3.2$
bash-3.2$ echo 'import {useDefault} from "react-use"' > a.mjs
bash-3.2$
bash-3.2$ node a.mjs
file:///private/tmp/a.mjs:1
import {useDefault} from "react-use"
        ^^^^^^^^^^
SyntaxError: Named export 'useDefault' not found. The requested module 'react-use' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from 'react-use';
const {useDefault} = pkg;

    at ModuleJob._instantiate (node:internal/modules/esm/module_job:128:21)
    at async ModuleJob.run (node:internal/modules/esm/module_job:194:5)
    at async Promise.all (index 0)
    at async ESMLoader.import (node:internal/modules/esm/loader:385:24)
    at async loadESM (node:internal/process/esm_loader:88:5)
    at async handleMainPromise (node:internal/modules/run_main:61:12)
```

`react-use` can be imported in a `.mjs` file with ESM syntax. 

**A little about versions:**  
- _OS_: macOS
- _Browser (vendor and version)_: N/A
- _React_: N/A
- _Node_: 16.15
- _`react-use`_: 17.4.0
- _Did this worked in the previous package version?_: No

oblador:
# Description

This package can't be used in a node environment using ESM as it's neither declared as `"type": "module"` nor does it include `.mjs` extensions for the files in the `module` main field. 

Since TSC is not able to produce proper ESM files with `.mjs` extensions I opted to use `tsup` to compile the ESM files, and I also had to add `@swc/core` to keep this backwards compatible with the `es5` target. I don't think there's much point in supporting back to es5 for an ESM, but I'm keeping it for now to avoid making this a breaking change. 

Fixes #2353

## Type of change

# Checklist
- [x] Add documentation

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,19 @@
   "version": "17.5.1",
   "description": "Collection of React Hooks",
   "main": "lib/index.js",
-  "module": "esm/index.js",
+  "module": "esm/index.mjs",
+  "exports": {
+    ".": {
+      "types": "./lib/index.d.ts",
+      "import": "./esm/index.mjs",
+      "require": "./lib/index.js"
+    },
+    "./*": {
+      "types": "./lib/*.d.ts",
+      "import": "./esm/*.mjs",
+      "require": "./lib/*.js"
+    }
+  },
   "sideEffects": false,
   "files": [
     "lib/",
@@ -21,7 +33,7 @@
     "lint:fix": "yarn lint --fix",
     "lint:types": "tsc --noEmit",
     "build:cjs": "tsc",
-    "build:es": "tsc -m esNext --outDir esm",
+    "build:es": "tsup --format esm --out-dir esm --clean src",
     "build": "yarn build:cjs && yarn build:es",
     "clean": "rimraf lib storybook-static esm",
     "storybook": "start-storybook -p 6008",
@@ -81,6 +93,7 @@
     "@storybook/addon-notes": "5.3.21",
     "@storybook/addon-options": "5.3.21",
     "@storybook/react": "6.4.9",
+    "@swc/core": "^1.7.26",
     "@testing-library/react": "12.1.2",
     "@testing-library/react-hooks": "7.0.2",
     "@types/jest": "27.5.2",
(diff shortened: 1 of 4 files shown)


bertho-zero:
@joecritch Can you look at this? It's been problematic for years

rxliuli:
The same question, is there a plan to merge this PR?

```js
Error in renderToPipeableStream: SyntaxError: [vite] Named export 'useThrottle' not found. The requested module 'react-use' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from 'react-use';
const {useThrottle} = pkg;
```

WaldoJeffers:
Any chance of getting this merged? I feel it's a simple change that would make a monumental difference

tomerzcod7:
How is this not yet supported? It really breaks SSRs frameworks

rxliuli:
> How is this not yet supported? It really breaks SSRs frameworks

The author seems to have abandoned maintaining this package, I'm considering whether to fork it. There hasn't been any meaningful commit throughout the entire 2025.

https://github.com/streamich/react-use/commits/master/

pamapa:
@streamich  are you still there? This MR looks pretty good and is highly needed

PR #2617, opened 2025-01-06
Remove unnecessary useEffect and ref in usePrevious hook

### LINKED ISSUE #2605: Use potentially unnecessary use of useEffect in usePrevious

The current implementation of the `usePrevious` hook relies on `useEffect` to store the previous state. While this approach works,  I believe that the use of useEffect might not be necessary. The usage of useEffect could be considered an anti-pattern according to the official React documentation (as described in [You Might Not Need an Effect](https://react.dev/learn/you-might-not-need-an-effect)) if I interpret it correctly. I think the `useEffect` in this hook adds unnecessary complexity and potentially increases rendering overhead.

It could be that more hooks such as `usePreviousDistinct` can be refactored without useEffect but I stumbled on this one first so I decided to put my focus this proposal first.

I propose an alternative implementation of `usePrevious` that eliminates the need for `useEffect` by using only `useRef`. This approach ensures that the hook is purely synchronous and avoids unnecessary side effects. Here's the proposed implementation:

```javascript
import { useRef } from "react";

export const usePrevious = <T>(state: T): T | undefined => {
    const ref = useRef<T>();

    const previousValue = ref.current;
    ref.current = state;

    return previousValue;
};
```
The implementation above updates the `ref` value synchronously with the component's render cycle, without requiring a separate side effect to handle state updates.

I was wondering if the implementation above could provide the same functionality but without using useEffect. It could be that I am missing edge cases but I'm more than happy to hear about them and add changes if possible.

None yet but I am open to investigate if necessary.

I would be happy to open a merge request with the proposed changes if this helps.

RickHuizing01:
# Description

This merge request resolves [GitHub Issue #2605](https://github.com/streamich/react-use/issues/2605) by updating the `usePrevious` hook to adhere to React's best practices. For a detailed explanation of the proposed implementation, please refer to the issue and its comments.

## Type of change

# Checklist
- [x] Add documentation

### DIFF
--- a/docs/usePrevious.md
+++ b/docs/usePrevious.md
@@ -1,6 +1,6 @@
 # `usePrevious`
 
-React state hook that returns the previous state as described in the [React hooks FAQ](https://reactjs.org/docs/hooks-faq.html#how-to-get-the-previous-props-or-state).
+React state hook that returns the previous state as described in [Adjusting some state when a prop changes](https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes).
 
 ## Usage
 
--- a/src/usePrevious.ts
+++ b/src/usePrevious.ts
@@ -1,11 +1,13 @@
-import { useEffect, useRef } from 'react';
+import { useState } from 'react';
 
 export default function usePrevious<T>(state: T): T | undefined {
-  const ref = useRef<T>();
+  const [current, setCurrent] = useState<T>(state);
+  const [previous, setPrevious] = useState<T>();
 
-  useEffect(() => {
-    ref.current = state;
-  });
+  if (current !== state) {
+    setCurrent(state);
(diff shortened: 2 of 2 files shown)


qwelol:
Hi! I also ran into a similar issue with my implementation of usePrevious.
It seems that the implementation introduced in this commit changes the behavior compared to the original.
Try running the test below and comparing the results — I suspect they'll differ.

```js
const setUp = () => renderHook(({ state }) => usePrevious(state), { initialProps: { state: 0 } });

it('should update value on rerender', () => {
  const { result } = setUp();

  expect(result.current).toBeUndefined();

  rerender({ state: 0 });

  expect(result.current).toBe(0); // 0 ? undefined ? What behavior is expected?
});
```

Is there any consensus on what the expected behavior should be in this case?
Let's discuss.

Issue #2501, opened 2023-06-03
[Feature Request] useMeasure() `ResizeObserver.observe()` option argument support


kkimdev:
ResizeObserver.observe() has `options` https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver/observe , but that's not configurable from `useMeasure()` currently.

My usecase is using `border-box` option.

arnaugomez:
I can take care of that. I will create a PR soon.

childrentime:
use this. https://www.reactuse.com/useMeasure

williamisnotdefined:
PR [here](https://github.com/streamich/react-use/pull/2619), feel free to help me if I forgot something.

PR #2619, opened 2025-01-24
[useMeasure] ResizeObserverOptions added as prop

williamisnotdefined:
# [useMeasure] ResizeObserverOptions added as prop

Now useMeasure accepts an `observerOptions` as parameter, plus the hook will also return the element ref in its last argument, just in case of some need.

This PR try to solve the issue [2501](https://github.com/streamich/react-use/issues/2501).
It is a really small change that shouldn't affect the current hook behavior if a parameter is not provided.
Almost sure I forgot something in the PR flow, please, just let me know.

## Type of change

# Checklist
- [x] Add documentation

### DIFF
--- a/docs/useMeasure.md
+++ b/docs/useMeasure.md
@@ -25,9 +25,32 @@ const Demo = () => {
 };
 ```
 
+## Usage with observer options
+```jsx
+import { useMeasure } from "react-use";
+
+const Demo = () => {
+  const [ref, { x, y, width, height, top, right, bottom, left }] = useMeasure({ observerOptions: { box: 'border-box' } });
+
+  return (
+    <div ref={ref}>
+      <div>x: {x}</div>	
+      <div>y: {y}</div>
+      <div>width: {width}</div>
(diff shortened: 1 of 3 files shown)


williamisnotdefined:
`yarn lint:types` returns the error 

> Cannot find name 'ResizeObserverOptions'

The resize-observer-polyfill does not support the `options` parameter in `observer` function:
```
declare global {
    interface ResizeObserverCallback {
        (entries: ResizeObserverEntry[], observer: ResizeObserver): void
    }

    interface ResizeObserverEntry {
        readonly target: Element;
        readonly contentRect: DOMRectReadOnly;
    }

    interface ResizeObserver {
        observe(target: Element): void; /* Here we should have the second parameter `options` */
        unobserve(target: Element): void;
        disconnect(): void;
    }
}
```

I think ResizeObserver is widely supported nowadays, source: [CanIUse](https://caniuse.com/resizeobserver)

image

Should we still rely upon the polyfill?

PR #2630, opened 2025-04-12
fix: ssr issue, move fns inside useEffect

awareness481:
# Description

The hook as is is not SSR safe (repro: [https://ck29z6-3000.csb.app/](https://ck29z6-3000.csb.app/) ). The timestamp is initialized on the server and on the client as `Date.now()` so it's guaranteed to cause a hydration issue when `.timestamp` is used on the DOM (unless the client's clock is behind and by chance it happens to be the same during client rendering).

Moreover, the server's clock is more than likely in a different timezone which would result in a wrong value being passed on the initial response.

Additionally, there's no point in initializing `.timestamp` as it arguably should only be initialized/set by the geolocation events.

Additional changes include: moving the functions inside useEffect so they're not needed in the deps array, removing `mounted` as it's a useless check. 

## Type of change

# Checklist
- [ ] Add documentation

### DIFF
--- a/src/useGeolocation.ts
+++ b/src/useGeolocation.ts
@@ -36,13 +36,11 @@ const useGeolocation = (options?: PositionOptions): GeoLocationSensorState => {
     latitude: null,
     longitude: null,
     speed: null,
-    timestamp: Date.now(),
+    timestamp: null,
   });
-  let mounted = true;
-  let watchId: any;
 
-  const onEvent = (event: any) => {
-    if (mounted) {
+  useEffect(() => {
+    const onEvent = (event: any) => {
       setState({
         loading: false,
         accuracy: event.coords.accuracy,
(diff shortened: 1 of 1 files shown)


src/useGeolocation.ts
```diff
@@ -54,17 +52,14 @@ const useGeolocation = (options?: PositionOptions): GeoLocationSensorState => {
...
-  useEffect(() => {
     navigator.geolocation.getCurrentPosition(onEvent, onEventError, options);
-    watchId = navigator.geolocation.watchPosition(onEvent, onEventError, options);
+    let watchId = navigator.geolocation.watchPosition(onEvent, onEventError, options);
 
     return () => {
-      mounted = false;
       navigator.geolocation.clearWatch(watchId);
     };
   }, []);
```
awareness481, line 65: Ideally, options should be added in the deps array but the changes the behavior of the hook

Issue #2614, opened 2024-12-19
Wrong file extension for esm


ardjuh:
**What is the current behavior?**
Vitest gives the following error when using a custom ui library with `react-use`:
`SyntaxError: Cannot use import statement outside a module
Module .../node_modules/ui-lib/.../node_modules/react-use/esm/useInterval.js:1 seems to be an ES Module but shipped in a CommonJS package. You might want to create an issue to the package "react-use" asking them to ship the file in .mjs extension or add "type": "module" in their package.json.

As a temporary workaround you can try to inline the package by updating your config:

// vitest.config.js
export default {
  test: {
    server: {
      deps: {
        inline: [
          "react-use"
        ]
      }
    }
  }
}`
This temporary workaround also doesn't fix the error.

**Steps to reproduce it and if possible a minimal demo of the problem. Your bug will get fixed much faster if we can run your code and it doesn't have extra dependencies other than `react-use`. Paste the link to your [JSFiddle](https://jsfiddle.net) or [CodeSandbox](https://codesandbox.io) example below:**
A ui library with
- react-use
- [rollup-plugin-tree-shakeable](https://github.com/TomerAberbach/rollup-plugin-tree-shakeable)
- The same React etc version

**What is the expected behavior?**
That the files in the esm folder have a .mjs extension and/or that the esm folder is marked is a module in the `package.json`

**A little about versions:**  
- _OS_: Not relevant
- _Browser (vendor and version)_: Not relevant
- _React_: 18.2.0
- _`react-use`_: 17.5.0
- _Did this worked in the previous package version?_: No

HimanshuKK08:
hey @ardjuh,  I'd love to give this a try as my first contribution. I understand the issue and will look into updating the ESM output configuration or adjusting the package structure. With a bit of guidance, I'm confident I can work through it. Let me know if it's okay for me to take this on!

ardjuh:
Hi @HimanshuKK08, yes please take on this issue.

HimanshuKK08:
hey @ardjuh, I have made changes to solve the issue, do checkout the PR

PR #2633, opened 2025-04-27
add 'type: module' field to package.json

HimanshuKK08:
This PR adds the "type": "module" field to the package.json file, which properly signals to Node.js and bundlers that the package uses ESM (ECMAScript Modules) syntax.

This change ensures better compatibility when consuming react-use in modern ESM environments, resolves issues related to module imports, and aligns with the ongoing transition to ESM support.

Fixes: #2614

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,6 +2,7 @@
   "name": "react-use",
   "version": "17.6.0",
   "description": "Collection of React Hooks",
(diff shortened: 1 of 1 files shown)


tjcchen (approved):

tjcchen:
Looks good!

PR #2640, opened 2025-07-03
fix(useLocalStorage): fix setter function returns incorrect previous value

kpCoder2801:
# Description

This PR fixes a bug in the `useLocalStorage` hook where the `setValue` function does not return the correct previous value. The issue occurs because the setter function is wrapped in a `useCallback`, but the current state value was not included in the dependency array. As a result, the stale value was being captured and returned incorrectly.

This fix adds the missing dependency to ensure the latest state is used in the callback.

## Type of change

# Checklist
- [ ] Add documentation

### DIFF
--- a/src/useLocalStorage.ts
+++ b/src/useLocalStorage.ts
@@ -19,13 +19,14 @@ const useLocalStorage = <T>(
   if (!isBrowser) {
     return [initialValue as T, noop, noop];
   }
+
   if (!key) {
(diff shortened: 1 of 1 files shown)


Redmega:
Without this useLocalStorage becomes highly unusable. Bumping for visibility -- would love if this could get merged.

I don't see any activity on the repo since December 2024, though.

PR #2705, opened 2026-06-07
feat: prefer screen.orientation change events in useOrientation

louzhedong:
## Summary
Prefer `screen.orientation.change` when available and fall back to `orientationchange`.

## Why
`orientationchange` is deprecated, while the Screen Orientation API exposes the modern event.

## What changed
- Updated the hook to subscribe to `screen.orientation.change` when possible
- Kept the legacy `orientationchange` fallback
- Added test coverage for both code paths

## Verification
- `jest tests/useOrientation.test.ts`
- `tsc --noEmit`

### DIFF
--- a/src/useOrientation.ts
+++ b/src/useOrientation.ts
@@ -16,6 +16,9 @@ const useOrientation = (initialState: OrientationState = defaultState) => {
 
   useEffect(() => {
     const screen = window.screen;
+    const orientation = screen.orientation as
+      | (ScreenOrientation & EventTarget)
+      | undefined;
     let mounted = true;
 
     const onChange = () => {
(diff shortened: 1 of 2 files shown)


louzhedong:
Pushed another update on top of this PR (`b88f063`) to keep the hook behavior aligned with the Screen Orientation API path.

Validation on the current patch:
- `eslint src/useOrientation.ts tests/useOrientation.test.ts`
- `prettier --check --single-quote src/useOrientation.ts tests/useOrientation.test.ts`
- `jest --runInBand --runTestsByPath tests/useOrientation.test.ts`

The branch is up to date and the tests are green locally.

PR #2712, opened 2026-06-27
fix(createReducer): return action from Dispatch to match Redux types

### LINKED ISSUE #856: useReducer not working with Typescript (type def errors showing up)
At the moment if you use useReducer as in the documentation you get weird errors when using Typescript:
image

It should work without complaining

**A little about versions:**  
- Typescript 3.5.3 - Typescript 3.7.4
- _OS_: Mojave
- _Browser (vendor and version)_: 
- _React_: 
- _`react-use`_: 13.12.2
- _Did this worked in the previous package version?_

mmustafasenoglu:
## Summary

The `Dispatch` type in `createReducer` was returning `void`, which caused TypeScript errors when using the hook with redux-thunk or other middleware that expect dispatch to return the action.

Fixes #856

## Changes

Changed `Dispatch<Action>` return type from `void` to `Action` to align with Redux's `Dispatch` typing.

```diff
- type Dispatch<Action> = (action: Action) => void;
+ type Dispatch<Action> = (action: Action) => Action;
```

## Testing

Verified that the type change matches Redux's dispatch signature and resolves the TypeScript compilation errors reported in the issue.


mmustafasenoglu:
Closing as the project seems inactive. Will reconsider if there's interest.

PR #2713, opened 2026-06-27
fix(useDeepCompareEffect): remove warning for primitive dependencies

### LINKED ISSUE #755: Remove warnings from `useDeepCompareEffect`

mmustafasenoglu:
## Summary

Remove the warning that fires when all dependencies are primitive values. This warning was incorrect because dependencies can change from primitive to non-primitive at runtime (e.g., `undefined` to an object), making `useDeepCompareEffect` the correct choice.

The 'no dependencies' warning is preserved as it indicates a likely mistake.

Fixes #755

## Changes

Removed the `deps.every(isPrimitive)` warning check from `useDeepCompareEffect`.

```diff
- if (deps.every(isPrimitive)) {
-   console.warn(
-     '`useDeepCompareEffect` should not be used with dependencies that are all primitive values. Use React.useEffect instead.'
-   );
- }
```

## Context

As noted by @streamich (owner): "I think the warnings for primitives in `useDeepCompareEffect` (and shallow compare) should be removed." The warning for no dependencies should stay.


mmustafasenoglu:
Closing as the project seems inactive. Will reconsider if there's interest.

PR #2720, opened 2026-09-22
feat: add useAllSearchParam hook

### LINKED ISSUE #2390: Add a useAllSearchParam hook
useSearchParam.ts only returns a single search param, however, I often want to get more than just a single URL search param and I don't want to keep instantiating the same hook for each time I want another param.

I would like a hook that returns all search params instead of just a single one

The only solution that I thought of was coping and pasting the hook on my project substituting the `.get` on line 4 for entries and using it to iterate all params and return an object

bosi-programming:
Adds a `useAllSearchParam` sensor hook that returns every URL search param as an object, so consumers don't need to instantiate `useSearchParam` once per param.

Closes #2390.

## Changes

- `src/useAllSearchParam.ts` — new hook returning `Record<string, string>`; watches `popstate`/`pushstate`/`replacestate`; server fallback returns `{}`.
- `tests/useAllSearchParam.test.ts` — covers all params, empty search, updates, duplicate keys.
- `src/index.ts` — exports the hook.
- `stories/useAllSearchParam.story.tsx`, `docs/useAllSearchParam.md`, `README.md` — docs and demo.

## Semantics

- Single value per key: a repeated param keeps its first value (`?foo=bar&foo=baz` → `{ foo: 'bar' }`), matching `useSearchParam`'s `.get()` behavior.

## Verification

- `yarn test` — 77 suites / 496 tests pass.
- `yarn lint` — 0 errors.
- `yarn lint:types` — clean.


bosi-programming:
Superseded by #2721 (branch renamed to `pr/use-all-search-param` per CONTRIBUTING.md, which closed this PR's head ref).

